PR #1, opened 2020-08-10
Custom Block System & Intellij stuff

hYdos:
A system that allows basic custom blocks to be added to the game. this pr also fixes a warning in IntelliJ IDEA in the Gradle wrapper.


hYdos:
haha yes also 1st pr

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
```diff
@@ -0,0 +1,92 @@
...
+    public boolean equals(Object obj) {
+        if(obj instanceof CustomBlock){
+            CustomBlock block = (CustomBlock) obj;
+            return block.id == this.id;
+        }else{
+            return super.equals(obj);
+        }
+    }
+
+    public static class Builder {
```
SuperCoder7979, line 62: Hmmm does it make sense here to use a builder? As there are only 2 params it might be easier to just pass them normally

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
```diff
@@ -0,0 +1,92 @@
...
+import net.minecraft.util.math.BlockPos;
+
+import javax.annotation.Nullable;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class CustomBlock {
+    private static final TinyRegistry<CustomBlock> REGISTRY = TinyRegistry.newStable();
+    private static final Map<BlockPos, CustomBlock> BLOCK_MAP = new HashMap<>();
```
SuperCoder7979, line 16: 1. A Long2ObjectOpenHashMap will boost the speed of the map here, as the block positions can be stored as long values.
2. A way to clear the map would be needed to prevent memory leaking.

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+        return false;
+    }
+
+    public boolean setBlock(BlockPos pos) {
+        return this.applyTo(pos);
+    }
+
+    @Nullable
+    public static CustomBlock match(BlockPos pos) {
```
SuperCoder7979, line 48:
```suggestion
    public static CustomBlock matches(BlockPos pos) {
```

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private final Identifier id;
+    private final Text name;
+
+    private CustomBlock(Identifier identifier, Text name) {
+        this.id = identifier;
+        this.name = name;
+    }
+
+    public static Set<Identifier> getKeys() {
```
SuperCoder7979, line 26:
```suggestion
    public static Set<Identifier> getIds() {
```

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
```diff
@@ -0,0 +1,92 @@
...
+        return this.applyTo(pos);
+    }
+
+    @Nullable
+    public static CustomBlock match(BlockPos pos) {
+        return BLOCK_MAP.get(pos);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
SuperCoder7979, line 53: Is the equals method needed here? It could also use some whitespace cleanup around the if and else statements.

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public boolean applyTo(BlockPos pos) {
+        if (this.name != null && !BLOCK_MAP.containsKey(pos)) {
+            BLOCK_MAP.put(pos, this);
+            return true;
+        }
+        return false;
+    }
+
+    public boolean setBlock(BlockPos pos) {
```
SuperCoder7979, line 43: Is this method needed? Seeing as it does the same thing as the above method 😅

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
hYdos: got it.

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
hYdos: there should only be 1 custom block at a pos so this name doesnt make sense

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
hYdos: not really. just a side effect of keeping as close to the custom item class as possible.

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
hYdos: if we are fixing that we should fix it in Custom Items

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
hYdos: Ive always wondered how much faster FastUtil maps are. how fast are they compared to normal java maps?

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
hYdos: in a future pr i plan to add more things such as enchantments, mining speed, callbacks, etc. im just preparing for it.

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
SuperCoder7979: They're much faster than normal maps and should always be used if possible.

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
SuperCoder7979: Alright, I guess that's fair

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
SuperCoder7979: Ah ok, it should be removed then

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
SuperCoder7979: Well yes, but you would say that `a block matches a position` and not `a block match a position`. It feels more correct that way.

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
SuperCoder7979: Hmm alright, put this change in and we'll change it there too

src/main/java/net/gegy1000/plasmid/shop/Cost.java
```diff
@@ -36,6 +36,18 @@ public static Cost ofEmeralds(int emeralds) {
                 .display(new LiteralText(emeralds + " emeralds").formatted(Formatting.GREEN));
     }
 
+    public static Cost free(){
+        return new Cost()
+                .take(Items.AIR, 0)
```
gegy1000, line 41: Maybe pass the `take` function and just return true

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
```diff
@@ -0,0 +1,97 @@
...
+import net.minecraft.util.Identifier;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+import javax.annotation.Nullable;
+import java.util.Set;
+
+public class CustomBlock {
+    private static final TinyRegistry<CustomBlock> REGISTRY = TinyRegistry.newStable();
+    private static final Long2ObjectArrayMap<CustomBlock> BLOCK_MAP = new Long2ObjectArrayMap<>();
```
gegy1000, line 18: This probably works for now, but I guess there's the worry that the block is updated in the world but the custom block isn't. Mostly notably being chunks being totally removed when games close

src/main/java/net/gegy1000/plasmid/shop/Cost.java
hYdos: what do you mean by that?

gegy1000: There's an overload to `take` which takes a function. Instead of trying to take an empty stack from the inventory, it could no-op directly and return successful (true)

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
hYdos: i might work on that in a future pr. i doubt anyone will run 2 games at once anyways. it crashes somewhere else if you do

src/main/java/net/gegy1000/plasmid/shop/Cost.java
hYdos: ah i see. sorry didnt see that

src/main/java/net/gegy1000/plasmid/item/CustomBlockItem.java
```diff
@@ -0,0 +1,58 @@
...
+        this.vanillaState = vanillaState;
+
+        UseBlockCallback.EVENT.register(this::onUse);
+    }
+
+    private ActionResult onUse(PlayerEntity playerEntity, World world, Hand hand, BlockHitResult blockHitResult) {
+        if(!world.isClient){
+            ItemStack stack = playerEntity.getStackInHand(hand);
+            CompoundTag tag = stack.getOrCreateTag();
+            if(tag.getString(Plasmid.ID + ":custom_item").equals(item.getIdentifier().toString())){
```
gegy1000, line 42: Might be possible to use the utils on CustomItem to match this

hYdos: fixed

hYdos:
ok all code is now formatted @SuperCoder7979. while im here should i cleanup the whole project using intellij?

src/main/java/net/gegy1000/plasmid/block/CustomBlock.java
```diff
@@ -0,0 +1,113 @@
...
+        if (this.name != null && !BLOCK_MAP.containsKey(pos.asLong())) {
+            BLOCK_MAP.put(pos.asLong(), this);
+            world.setBlockState(pos, state);
+            return true;
+        }
+        return false;
+    }
+
+    @Nullable
+    public static CustomBlock matches(BlockPos pos) {
```
gegy1000, line 75: This should probably just be `match` or `get`

hYdos: *unhappy supercoder sounds*. ill just go back to match() like items

gegy1000: 😳

hYdos: is that all the changes you want done?

PR #7, opened 2020-08-12
Add the ability to fake registry entries

ramidzkh:
Add the ability to fake blocks, items and potentially others with the client

Known issues:
- [X] ~~Custom blocks get sent as air (I have a WIP mildly-invasive fix)~~


Restioson:
What does "fake blocks" mean?

ramidzkh:
> What does "fake blocks, items" mean?

The server logically uses Items and Blocks, but data sent to the client uses something it can understand, such as replacing a modded item with a diamond.

src/main/java/xyz/nucleoid/plasmid/mixin/fake/IdListMixin.java
```diff
@@ -0,0 +1,48 @@
...
+        if (entry instanceof BlockState) {
+            BlockState state = (BlockState) entry;
+            Block block = state.getBlock();
+
+            if (block instanceof FakeBlock) {
+                return (T) ((FakeBlock<?>) block).getFaking(state);
+            }
+        }
+
+        if (entry instanceof ItemStack) {
```
gegy1000, line 33: Is `IdList` ever used for `ItemStacks`?

gegy1000:
Will the server try to sync the registry entries for the fake blocks/items?

src/main/java/xyz/nucleoid/plasmid/mixin/fake/IdListMixin.java
ramidzkh: No but other mods may have an IdList for ItemStacks, ¯\\_(ツ)_/¯

ramidzkh:
I've tested this with a client with only Fabric API installed, and it works like a charm

src/main/java/xyz/nucleoid/plasmid/mixin/fake/IdListMixin.java
gegy1000: Would that even be possible? 😁 there’s no registry for ItemStacks like there is for BlockState

gegy1000:
Not super familiar with how the handshake goes. I’m sure it’s fine if it’s working but I wonder if it is still syncing, and if that might have any unforeseen implications?

oliviathevampire:
what would this allow for?

Restioson:
AFAIK it would let you add a modded item presented to the player as a non modded item - e.g a gun presented as a hoe.

ramidzkh:
image
The 'brick' is actually a custom block with a direction property. The fluid is probably not flowing since I rushed the testing fluid implementation. I forgot to include items in the screenshot, but they do work. Also persists save and such

PR #10, opened 2020-08-12
Update Player death event with javadocs + ActionResult instead of bools

Draylar:
PR that updates the Player Death Event with JavaDocs, and makes it use ActionResult (which makes it much more intuitive/useable).

I also noticed some of the listeners could be shortened with lambdas (which I did here). Is this intentional, or am I fine condensing it down?


gegy1000:
Nice :D
Should we do the same for the damage event in this PR, or do it separately?

I only put the lambdas on separate lines because I thought it looked nicer- but it doesn't _really_ matter :P

Draylar:
I was planning on doing that (updates to other events) in a separate PR after I got feedback on this one.

PR #11, opened 2020-08-12
Add a way to track the ConfiguredGame of a GameWorld.

arthurbambou:


src/main/java/xyz/nucleoid/plasmid/game/config/GameConfigs.java
```diff
@@ -80,6 +80,15 @@ private static Identifier identifierFromPath(Identifier location) {
         return CONFIGURED_GAMES.get(identifier);
     }
 
+    @Nullable
+    public static Identifier getID(ConfiguredGame<?> configuredGame) {
```
SuperCoder7979, line 84:
```suggestion
    public static Identifier getId(ConfiguredGame<?> configuredGame) {
```

src/main/java/xyz/nucleoid/plasmid/game/ConfiguredGame.java
```diff
@@ -4,23 +4,31 @@
...
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
 public final class ConfiguredGame<C> {
     public static final Codec<ConfiguredGame<?>> CODEC = GameType.REGISTRY.dispatchStable(c -> c.type, ConfiguredGame::codecFor);
 
     private final GameType<C> type;
     private final C config;
+    private static List<ConfiguredGame<?>> openingGames = new ArrayList<>();
```
SuperCoder7979, line 16: Is there a reason why this is a list? A non-list based approach may be easier here 😅

arthurbambou: If two game opens before the GameWorld is open, the ConfiguredGame will not match the GameWorld.

SuperCoder7979: That's true- but a list doesn't seem like the most appropriate solution in this instance.

arthurbambou: What would you suggest then?

SuperCoder7979: A usecase for this feature would probably help in making that decision 😅

arthurbambou: I wanted to show more information in game-gui when choosing a game to join. And to get those info I need to be able to know the ConfiguratedGame being played in a GameWorld.

arthurbambou: What it looks like rn. image

gegy1000:
I think the approach to take here is pass the game initializers an object that creates GameWorlds under the specific config, so the game initializer never has to know what config is running it

arthurbambou:
> so the game initializer never has to know what config is running it
What do you mean?

gegy1000:
Sorry, didn't explain so great. Just pushed a commit that implements it: https://github.com/NucleoidMC/plasmid/commit/e5758dc797fc2143a6edc281e57fcfd3759dac12

arthurbambou:
awesome! I gonna close this then.

PR #12, opened 2020-08-12
Add GameWorld Javadocs + some listener refactoring/jdocs + GameWorld refactoring

Draylar:
This PR adds Javadocs to all `GameWorld` methods, refactors some listeners to use `ActionResult`, and cleans up `GameWorld`.

The `GameWorld` refactor gives `GameWorld` a `Game` instance instead of the `EventListeners` and `GameRuleSet` field it already had (`Game` exists to represent those two fields already, so there's no point duplicating them). It also removes the `ServerWorld` instance in favor of using the one in `BubbleWorld` (this means you only have to track/update 1, plus it solidifies what `BubbleWorld` actually is).

Let me know if any of the method docs don't make sense (or if I misunderstood/misrepresented something).


Draylar:
The latest commit has new Javadocs for `StartResult`. I also slightly refactored the template errors to be static fields instead of static methods (this means you're not creating a new error *every time* you want it, + makes it easier to understand how to use the class).

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -94,33 +130,65 @@ public void setGame(Game game) {
         this.invoker(GameOpenListener.EVENT).onOpen();
     }
 
+    /**
+     * Closes this {@link GameWorld}.
```
gegy1000, line 134: This closes specifically the current Game instance stored and notifies it, but doesn't close the world itself

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -94,33 +130,65 @@ public void setGame(Game game) {
...
+     */
     private void closeGame() {
         this.invoker(GameCloseListener.EVENT).onClose();
-        this.listeners = new EventListeners();
-        this.rules = GameRuleSet.empty();
+        this.game = new Game();
     }
 
+    /**
+     * Attempts to add the given {@link ServerPlayerEntity} to this {@link GameWorld}.
```
gegy1000, line 144: I think also worth not mentioning bubble worlds here and rather describing the function: a player *must* be added by this function, or they cannot exist in the world. Also can describe that it does not generally need to be manually called, because it's already controlled by the player offer abstraction

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -153,10 +234,24 @@ public void tick() {
         this.invoker(GameTickListener.EVENT).onTick();
     }
 
+    /**
+     * Requests that this {@link GameWorld} begins, which invokes any {@link RequestStartListener#EVENT} listener attached to it.
```
gegy1000, line 238: Maybe can also mention that the meaning of a game being "started" is up to interpretation of the game implementation and that it's specifically connected to `/game start`

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -170,6 +265,11 @@ public JoinResult offerPlayer(ServerPlayerEntity player) {
         }
     }
 
+    /**
+     * Closes this {@link GameWorld} if it is not already closed.
+     *
+     * <p>Upon close, all players in this {@link GameWorld} are removed (kick implementation is up to the {@link BubbleWorld} associated with this {@link GameWorld}).
```
gegy1000, line 271: Without the mention of bubble world, could just say they are kicked back and restored to their previous state former to joining

src/main/java/xyz/nucleoid/plasmid/game/StartResult.java
```diff
@@ -6,38 +6,64 @@
...
     private final Text error;
 
     private StartResult(Text error) {
         this.error = error;
     }
 
+    /**
+     * Returns a {@link StartResult} with no error.
+     *
+     * @return {@link StartResult} with no error (null)
```
gegy1000, line 25: I think the implementation of using a null error should be hidden- this just represents a successful result

src/main/java/xyz/nucleoid/plasmid/game/StartResult.java
```diff
@@ -6,38 +6,64 @@
 
 import javax.annotation.Nullable;
 
+/**
+ * Describes the result of a {@link GameWorld} attempting to start.
+ */
 public final class StartResult {
+    public static final StartResult ALREADY_STARTED = error(new LiteralText("This game has already started!"));
```
gegy1000, line 13: It's a stylistic thing, but personally prefer to keep the consistency of having them all fields/all methods.

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -31,22 +23,38 @@
...
+    private Game game = new Game();
 
     private boolean closed;
 
+    /**
+     * Primary constructor for {@link GameWorld}.
+     *
+     * @param bubble the {@link BubbleWorld} this {@link GameWorld} is hosted in.
+     */
     private GameWorld(BubbleWorld bubble) {
```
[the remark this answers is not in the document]
hYdos: better having a doc then not having one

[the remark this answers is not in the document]
gegy1000: I’d disagree- it’s a private method, and it does nothing but setting fields. It’s just extra noise, and you don’t want to make users think they can access this method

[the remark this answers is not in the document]
hYdos: oh i didnt see the private part of that field. misunderstood it sorry

src/main/java/xyz/nucleoid/plasmid/command/GameCommand.java
```diff
@@ -181,7 +181,7 @@ private static int startGame(CommandContext<ServerCommandSource> context) throws
         CompletableFuture<StartResult> resultFuture = CompletableFuture.supplyAsync(gameWorld::requestStart, server);
 
         resultFuture.thenAccept(startResult -> {
-            if (startResult.isErr()) {
+            if (startResult.hasError()) {
```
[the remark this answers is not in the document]
Draylar: Fair. SuperCoder said it was named `err` for a reason related to Rust, do you want to keep that, or is `isError` fine? I think `isError` makes more sense (I originally thought whoever wrote the method name was just trying to save time, lol).

src/main/java/xyz/nucleoid/plasmid/game/Game.java
```diff
@@ -20,11 +25,21 @@ public Game setRule(GameRule rule, RuleResult result) {
         return this;
     }
 
+    /**
+     * Returns the {@link EventListeners} associated with this {@link Game}.
+     *
+     * @return the {@link EventListeners} of this {@link Game}
```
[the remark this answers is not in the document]
Draylar: I fixed this instance and a couple of other instances like it. Let me know if I missed any.

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
Draylar: Fixed. Any thoughts on whether docs should tell the caller about notifying listeners (eg. `"All XYZ listeners are notified of this action."`)?

src/main/java/xyz/nucleoid/plasmid/game/StartResult.java
Draylar: Are you saying I should keep the fields private and offer methods for returning them (so similar to before, but without an object being created on every call)?

src/main/java/xyz/nucleoid/plasmid/command/GameCommand.java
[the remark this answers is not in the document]
gegy1000: `isError` is fine - there's no specific reason to use `err` it's just a convention I'm used to :P

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
gegy1000: Oh, yeah. That would be nice to document. :)

src/main/java/xyz/nucleoid/plasmid/game/StartResult.java
gegy1000: Yea- either that or can make `OK` a field also

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -31,24 +23,37 @@
 import java.util.Set;
 import java.util.function.Consumer;
 
+/**
+ * Represents a unique world with a {@link Game} attached to it.
+ *
+ * <p>Each world has a generator which is triggered when the world is first created.
```
gegy1000, line 29:
```suggestion
 * <p>Each world has a {@link ChunkGenerator} which is invoked as chunks are requested.
 It is important to note that not all chunks will be loaded on start, and the game logic must take care to handle this.
```

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -100,33 +133,63 @@ public void setGame(Game game) {
...
+     */
     private void closeGame() {
         this.invoker(GameCloseListener.EVENT).onClose();
-        this.listeners = new EventListeners();
-        this.rules = GameRuleSet.empty();
+        this.game = new Game();
     }
 
+    /**
+     * Attempts to add the given {@link ServerPlayerEntity} to this game world.
```
gegy1000, line 147:
```suggestion
     * Attempts to add the given {@link ServerPlayerEntity} to this game world if it is not already added.
```

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -100,33 +133,63 @@ public void setGame(Game game) {
...
+     *
+     * @param player {@link ServerPlayerEntity} to add to this {@link GameWorld}
+     * @return whether the {@link ServerPlayerEntity} was successfully added
+     */
     public boolean addPlayer(ServerPlayerEntity player) {
         return this.bubble.addPlayer(player);
     }
 
+    /**
+     * Attempts to removes the given {@link ServerPlayerEntity} from this {@link GameWorld}.
```
gegy1000, line 159:
```suggestion
     * Attempts to removes the given {@link ServerPlayerEntity} from this {@link GameWorld}.
     * When a player is removed, they will be teleported back to their former location prior to joining
```

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -144,13 +207,26 @@ public boolean containsEntity(LivingEntity entity) {
...
+     */
     public RuleResult testRule(GameRule rule) {
-        return this.rules.test(rule);
+        return this.game.getRules().test(rule);
     }
 
+    /**
+     * Attempts to tick this {@link GameWorld} and any {@link GameTickListener#EVENT} listener attached to it.
+     *
+     * <p>If this {@link GameWorld} is closed ({@link GameWorld#closed}), the tick will not execute.
```
gegy1000, line 228: I think it's not important to mention the internal field :)
```suggestion
     * <p>If this {@link GameWorld} has been closed, the tick will not execute.
```

src/main/java/xyz/nucleoid/plasmid/game/StartResult.java
```diff
@@ -6,38 +6,62 @@
...
-    }
-
-    public static StartResult alreadyStarted() {
-        return err(new LiteralText("This game has already started!"));
-    }
-
+    /**
+     * Returns whether this {@link StartResult} is a success (no error).
+     *
+     * @return whether this {@link StartResult} is a success (no error).
```
gegy1000, line 43: Can remove the duplicated documentation here and just have the `@return` line

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
Draylar: Thanks. My initial intention was to show how to check whether the world is closed already, but the field is private anyway, so it doesn't matter.

src/main/java/xyz/nucleoid/plasmid/game/StartResult.java
Draylar: I think fields make more sense, only because it helps to suggest (while autocompleting) that you can also make your own (and that the fields are just templates). I think I'm going to refactor `JoinResult`/`StartResult` in a separate PR later anyways, so I left `JoinResult` as-is for now.

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
Draylar: Thanks. Part of me thinks I should change it to just be `Adds the given {@link ServerPlayerEntity} to this game world if it/they is not already added/present.`, but I think your version works for now.

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
Draylar: Sounds good, just wanted to make sure I should leave it in. All methods that call listeners in `GameWorld` mention it in one way or another now, and I'll continue that when I document other classes.

PR #13, opened 2020-08-13
fix custom items not working after inventory is opened.

arthurbambou:
fix custom items not working after inventory is opened.


hYdos:
i didnt even get time to approve this

PR #16, opened 2020-08-13
Force packets to be flattened to the integrated server

ramidzkh:
Naive way for now. It works well enough


gegy1000:
So.. this is definitely *very* hacky, lol. Any idea what other ways of approaching this would look like? Are they worth considering or would they be too invasive?

ramidzkh:
I don't think going through all the chunk delta/data packets and reflectively scanning through all packets (remember, mods!) to replace ItemStacks would be nice

gegy1000:
Fair enough :D

PR #19, opened 2020-08-14
Added chat channels

Restioson:
I need more than one person to test :pensive:


src/main/java/xyz/nucleoid/plasmid/command/ChatCommand.java
```diff
@@ -0,0 +1,33 @@
...
+        dispatcher.register(
+            literal("chat")
+                .then(literal("all").executes(ChatCommand::switchToAll))
+                .then(literal("team").executes(ChatCommand::switchToTeam))
+        );
+    }
+    // @formatter:on
+
+    public static int switchToAll(CommandContext<ServerCommandSource> ctx) throws CommandSyntaxException {
+        ((HasChatChannel) ctx.getSource().getPlayer()).setChatChannel(ChatChannel.ALL);
```
gegy1000, line 25: These should probably send feedback when running the commands to let you know they've done something

PR #30, opened 2020-08-18
Game Channels

gegy1000:
This introduces the concept of "game channels" to Plasmid. A channel is a collection of endpoints (such as a sign or entity) which control access into a specific game. The channel can control how it wants to display metadata about itself, and is responsible for updating that display whenever the game state changes.

To accomplish this, some refactoring was needed. This is mainly in the ability for outside listeners to subscribe to lifecycle events on the `GameWorld`. Additionally, have done some refactoring such that game events should *always* be called from the main tick thread in order to avoid some issues such as inventory desync. This means that `GameWorld.open` now returns a future instead of returning immediately, which is a breaking change.

Have also made some package changes as some have been expanding very quickly, and we're making breaking changes here already.

[Video displaying channel function](https://cdn.discordapp.com/attachments/733463265341866046/745273948320563240/2020-08-18_15-31-44.mp4)

Currently, not totally sure about design. Any feedback is appreciated here. Also important to consider how things like Game-Gui should hook into this system. I'm imagining that it could potentially add a custom game channel type, but we'd need to look deeper into the requirements that would have. Most notably for now is probably how non-default channel types should be created. Currently, this is done through `/game channel open <id> <game_id>`, but for a custom channel type, we'd have to figure something else out.

Another opportunity for a new game channel implementation is through matchmaking: a matchmaking channel type would abstract over many game instances and choose which to add a player to.


src/main/java/xyz/nucleoid/plasmid/mixin/game/event/ProjectileEntityMixin.java
```diff
@@ -0,0 +1,45 @@
...
+import xyz.nucleoid.plasmid.game.event.EntityHitListener;
+
+@Mixin({ ProjectileEntity.class })
+public abstract class ProjectileEntityMixin extends Entity {
+    public ProjectileEntityMixin(EntityType<?> type, World world) {
+        super(type, world);
+    }
+
+    @Inject(method = "onCollision", at = @At("HEAD"), cancellable = true)
+    private void onCollision(HitResult hitResult, CallbackInfo ci) {
```
SmellyModder, line 26: Perhaps it would be better to have only one event get invoked here? Just a `ProjectileHitListener`.
It'd make the code in this method cleaner and possibly add more freedom to the listeners, however, I understand why you went with the two events approach so that games could target specifically what they need.

src/main/java/xyz/nucleoid/plasmid/entity/CustomEntity.java
```diff
@@ -13,6 +13,7 @@
 
 import javax.annotation.Nullable;
 
+@Deprecated
```
SuperCoder7979, line 16: Some deprecation comments would be nice here, so devs know what to use instead :D

src/main/java/xyz/nucleoid/plasmid/entity/CustomizableEntity.java
```diff
@@ -2,6 +2,7 @@
 
 import javax.annotation.Nullable;
 
+@Deprecated
```
SuperCoder7979, line 5: same here :P

src/main/java/xyz/nucleoid/plasmid/command/GameCommand.java
```diff
@@ -235,12 +258,74 @@ private static int listGames(CommandContext<ServerCommandSource> context) {
...
     }
 
-    private static SuggestionProvider<ServerCommandSource> gameSuggestions() {
-        return (ctx, builder) -> {
-            return CommandSource.suggestMatching(
-                    GameConfigs.getKeys().stream().map(Identifier::toString),
-                    builder
-            );
-        };
+    private static int openChannel(CommandContext<ServerCommandSource> context) throws CommandSyntaxException {
```
SmellyModder, line 261: `context.getSource()` is called twice in here, so perhaps it'd be better to have it be a local variable?
This nitpick also applies for `removeChannel()` and `connectBlockToChannel()`

src/main/java/xyz/nucleoid/plasmid/game/GameWorld.java
```diff
@@ -153,20 +161,33 @@ private void closeGame() {
...
-        return resource;
+        synchronized (this.resources) {
+            this.resources.add(resource);
+            return resource;
+        }
     }
 
     /**
      * Attempts to add the given {@link ServerPlayerEntity} to this game world if it is not already added.
+     * This should be
```
SuperCoder7979, line 172: I think you forgot to finish the comment here 😅

src/main/java/xyz/nucleoid/plasmid/mixin/game/event/ProjectileEntityMixin.java
gegy1000: Hmm, good point. Also, weird, not sure why git is showing this as a new file- this was added in a previous commit. @SuperCoder7979, thoughts on changing this?

src/main/java/xyz/nucleoid/plasmid/world/bubble/HasBubbleWorld.java
```diff
@@ -0,0 +1,10 @@
+package xyz.nucleoid.plasmid.world.bubble;
+
+import javax.annotation.Nullable;
+
+public interface HasBubbleWorld {
```
SuperCoder7979, line 5: hmmm maybe this should be named `BubbleWorldHolder` or something? the "Has" prefix seems a bit weird to me

src/main/java/xyz/nucleoid/plasmid/mixin/game/event/ProjectileEntityMixin.java
```diff
@@ -0,0 +1,45 @@
...
+import net.minecraft.world.World;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import xyz.nucleoid.plasmid.game.GameWorld;
+import xyz.nucleoid.plasmid.game.event.BlockHitListener;
+import xyz.nucleoid.plasmid.game.event.EntityHitListener;
+
+@Mixin({ ProjectileEntity.class })
```
SuperCoder7979, line 19: oh oops- I forgot to remove the {} here, this is a note so i remember to fix that :P

src/main/java/xyz/nucleoid/plasmid/world/bubble/BubbleWorldSpawner.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.util.math.Vec3d;
+import net.minecraft.world.Heightmap;
+import net.minecraft.world.chunk.Chunk;
+
+public interface BubbleWorldSpawner {
+    static BubbleWorldSpawner at(Vec3d pos) {
+        return new At(pos);
+    }
+
+    static BubbleWorldSpawner atSurface(BlockPos pos) {
```
SuperCoder7979, line 15: I think an `atSurface(int x, int z)` would be useful in addition to the blockpos version, so we don't need to construct it if we don't have to

src/main/java/xyz/nucleoid/plasmid/mixin/game/event/ProjectileEntityMixin.java
SuperCoder7979: I thought that using 2 listeners would allow for better code, as with a singular hit listener you would need extra logic for switching between the type and casting, which would be better if it was abstracted away, in my opinion.

SmellyModder: That was one of my thoughts when looking it over yeah. Mostly was just looking to see what people would prefer 😛

PR #37, opened 2020-08-26
Added ServerData - A simple way to store data on servers

SmellyModder:
The title says it all basically.

This adds an easy and simple way to store data on servers.
This is very useful for games as it can be used to keep track of player scores, player ranks, player levels, etc.

I've also included the `PlayerScoreStorage` class so that games can store the scores of players easier.


src/main/java/xyz/nucleoid/plasmid/mixin/bubble/ServerWorldMixin.java
```diff
@@ -61,6 +66,13 @@ private ServerWorldMixin(MutableWorldProperties properties, RegistryKey<World> r
         super(properties, registryKey, dimension, profiler, client, debug, seed);
     }
 
+    @Inject(method = "<init>", at = @At("RETURN"))
```
gegy1000, line 69: We should probably add a new mixin class for this in a different package - it should be fine to have multiple ServerWorldMixins.

src/main/java/xyz/nucleoid/plasmid/storage/PlayerScoreStorage.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * A simple class for storing the scores of players.
+ * 'Score' can mean any number.
+ */
+public class PlayerScoreStorage implements ServerStorage {
+    public final Map<UUID, Float> scoreMap = new HashMap<>();
```
gegy1000, line 18: This can use an `Object2Float` map

gegy1000 (commented):
This is looking great! Just have some minor nitpicks

Restioson:
How is the data stored on disk?

ramidzkh:
Using PersistentStateManager on the overworld

ramidzkh (approved):

Restioson:
Ah okay, cool! So it is K/V store or what?

src/main/java/xyz/nucleoid/plasmid/mixin/bubble/ServerWorldMixin.java
gegy1000: On second thought, this might not be necessary at all: I don't think it's strictly required that the `PersistentState` is initialized on world init

src/main/java/xyz/nucleoid/plasmid/storage/PlayerScoreStorage.java
gegy1000: It might also make sense to use an int just for precision-loss purposes?

src/main/java/xyz/nucleoid/plasmid/mixin/bubble/ServerWorldMixin.java
SmellyModder: Yes, a `PersistentState` doesn't have to be initialized on world init. 
It only reads the file when the `PersistentState` is first created. 
But where do you think it might be a good place to do it since it still needs the Overworld `ServerWorld`?.
A possible area I can think of would be `MinecraftServer#loadWorld()`

src/main/java/xyz/nucleoid/plasmid/storage/PlayerScoreStorage.java
SmellyModder: I made it not use an int since some games may want to store scores that would be decimals, such as how fast you completed a game.
Although yes the precision-loss could be a problem, so I could make it use double instead.

src/main/java/xyz/nucleoid/plasmid/mixin/bubble/ServerWorldMixin.java
gegy1000: The initialization logic can just be delegated to getOrCreate- the get function could go through the MinecraftServer to get the overworld instance

src/main/java/xyz/nucleoid/plasmid/storage/PlayerScoreStorage.java
gegy1000: Double would be better, but it's still not possible to guarantee total accuracy. I think if something wants to store decimal points they could store with some constant denominator?

src/main/java/xyz/nucleoid/plasmid/mixin/bubble/ServerWorldMixin.java
SmellyModder: Oh yeah sure, but where should I call the `get()` method?
It needs to happen sometime after `MinecraftServer#createWorlds()` so it has the Overworld.

src/main/java/xyz/nucleoid/plasmid/storage/PlayerScoreStorage.java
SmellyModder: I suppose someone could just store it with a constant denominator and could always just make their own score storage.
I'll make it use an `Object2IntMap` instead then 👍

src/main/java/xyz/nucleoid/plasmid/mixin/bubble/ServerWorldMixin.java
gegy1000: The `get()` function doesn't ever need to be called by plasmid- whatever code is making use of the storage can call it whenever it needs to and it can initialize then

SmellyModder: I suppose it could be made to work like that, but if that is how it'd work then mods would need to call `ServerStorageManager#get()` every time they plan on loading their `ServerStorage`(s) which could get repetitive and a bit ugly looking and would also require `MinecraftServer` access. 

The reason I wanted to initialize it when the Server is loading is so that it's cleaner and easier for mods to load their `ServerStorage`(s) since it would load all the `ServerData`s in a bunch.

gegy1000: oh! duh, okay- i get it. sorry about that :P
this should be good to just have in a separate mixin file then 👍

SuperCoder7979:
Looks cool! This will be super useful to have for a bunch of minigames 😄

SmellyModder:
Yep, it's gonna be very helpful 🙂

src/main/java/xyz/nucleoid/plasmid/storage/ServerStorage.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Adds a {@link ServerStorage} to the {@link #STORAGES} map and returns the {@link ServerStorage}.
+     * Use this to have your {@link ServerStorage} be saved and loaded.
+     * @param key - The id of the storage.
+     * @param storage - The {@link ServerStorage} to add to the {@link #STORAGES} map.
+     * @param <S> - The type of {@link ServerStorage}.
+     * @return - The supplied {@link ServerStorage}.
+     */
+    static <S extends ServerStorage> S createStorage(Identifier key, S storage) {
```
gegy1000, line 26: Should this be allowed to be called after world load? I guess it should be fine under the circumstance that nothing has already been saved for that key, but it'd leave the storage to never be loaded

SmellyModder: Hm, yeah fair point. Perhaps it should throw an exception if it's called after the storages have been loaded?

gegy1000: Yeah, that'd work 🙂

gegy1000: Alternatively- although, this may be a little more annoying to use, we could force instances to be registered through the PersistentState object

SmellyModder: Yeah, fair.
But preferably this system should be very simple and easy to use so just throwing an exception works good in my opinion, and I don't imagine people trying to create new storages after it loads since it wouldn't load that storage at all and the exception works as a good notifier and catch for that.

gegy1000: Yeah, thinking about it more, that's definitely the better solution :)

gegy1000 (approved):

PR #38, opened 2020-08-26
Add support for Server-Translations - Add a way to send translated message to player in their client language.

arthurbambou:
This add an easy way to send translated message when Server-Translations is on the server.
If Server-Translations is not on the server, it will send a default message.

There is probably a better way to do it, feel free to suggest any idea ^^.

Tested outside of dev environment.


src/main/java/xyz/nucleoid/plasmid/command/GameCommand.java
```diff
@@ -210,8 +211,10 @@ private static int startGame(CommandContext<ServerCommandSource> context) throws
                 Text error = startResult.getError();
                 message = error.shallowCopy().formatted(Formatting.RED);
             } else {
-                message = source.getDisplayName().shallowCopy().append(" has started the game!")
-                        .formatted(Formatting.GRAY);
+                message = new TranslatableLiteralText(source.getDisplayName().shallowCopy().append(" has started the game!"),
```
gegy1000, line 214: Hmm- would it be possible to have the messages not duplicated?

arthurbambou: It's the default message used when Server-Translation is not on the server.

arthurbambou: It depends if you prefer Server-Translations to be a hard dependency or a soft one.

gegy1000: Ah okay, I understand. Might it make sense to have the mod be JIJ'd? Or if not JIJ'd completely, a small part of the API that delegates to en_us translation if not installed?

arthurbambou: Opened #39 for the not JIJ'd completely case.

PR #39, opened 2020-08-26
Add support for Server-Translations - Soft dependent version

arthurbambou:
#38 remade using Gegy's suggestions in case of soft dependent.


arthurbambou:
Ready for review

PR #42, opened 2020-08-29
Make scoreboard translatable

arthurbambou:


src/main/java/xyz/nucleoid/plasmid/widget/SidebarWidget.java
```diff
@@ -40,7 +46,10 @@ public static SidebarWidget open(Text title, PlayerSet players) {
         return widget;
     }
 
-    public void set(String[] display) {
+    /*
+       Style wont work.
+     */
+    public void set(TranslatableText[] display) {
```
gegy1000, line 52: I’m not so sure about how we’ll want to set this out. I think it might make sense to somehow use our own text-representing object?

Restioson:
Why the close?

LambdAurora:
Targeted branch got yeeted so it can't be merged into a non-existent branch

PR #47, opened 2020-09-07
Add additional arbitrary data to regions and new commands.

LambdAurora:
This is very noice.

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/command/MapCommand.java
+++ b/src/main/java/xyz/nucleoid/plasmid/command/MapCommand.java
@@ -11,7 +11,6 @@ import com.mojang.brigadier.suggestion.SuggestionProvider;
 import net.minecraft.command.argument.BlockPosArgumentType;
 import net.minecraft.command.argument.IdentifierArgumentType;
 import net.minecraft.command.argument.NbtCompoundTagArgumentType;
-import net.minecraft.data.client.model.BlockStateVariantMap;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.server.command.CommandSource;
 import net.minecraft.server.command.ServerCommandSource;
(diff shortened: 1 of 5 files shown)


src/main/java/xyz/nucleoid/plasmid/game/map/template/TemplateRegion.java
```diff
@@ -20,14 +23,25 @@ public BlockBounds getBounds() {
...
+     * @return The extra data.
+     */
+    public @Nullable CompoundTag getData() {
+        return this.data;
+    }
+
     public CompoundTag serialize(CompoundTag tag) {
         tag.putString("marker", this.marker);
         this.bounds.serialize(tag);
+        if (this.data == null) { tag.put("data", new CompoundTag()); } else { tag.put("data", this.data); }
```
gegy1000, line 38: It's not super significant, but there's some inconsistency in that a data tag may be null when created, and then later become an empty compound tag when loaded from disk. It might make sense to make it stored as nullable in nbt with a `contains` check on deserialization, or it might make sense to just have the data value be nonnull and default initialized to empty tag.

LambdAurora: I think data should not be nullable, so new commit to fix that!

src/main/java/xyz/nucleoid/plasmid/command/MapCommand.java
```diff
@@ -193,6 +224,74 @@ private static int addRegion(CommandContext<ServerCommandSource> context) throws
         return Command.SINGLE_SUCCESS;
     }
 
+    private static int renameRegions(CommandContext<ServerCommandSource> context, BlockStateVariantMap.TriFunction<TemplateRegion, String, BlockPos, Boolean> predicate) throws CommandSyntaxException {
```
gegy1000, line 227: Might make sense to use a custom interface for the predicate- that is one real chunky boi :D

src/main/java/xyz/nucleoid/plasmid/command/MapCommand.java
```diff
@@ -193,6 +224,74 @@ private static int addRegion(CommandContext<ServerCommandSource> context) throws
...
+        ServerCommandSource source = context.getSource();
+        ServerWorld world = source.getWorld();
+        BlockPos pos = source.getPlayer().getBlockPos();
+
+        StagingMapManager stagingMapManager = StagingMapManager.get(world);
+
+        String oldMarker = StringArgumentType.getString(context, "old");
+        String newMarker = StringArgumentType.getString(context, "new");
+
+        Optional<StagingMapTemplate> mapOpt = stagingMapManager.getStagingMaps().stream()
```
gegy1000, line 237: Would it be possible to extract this common code for finding staging map & templates?

src/main/java/xyz/nucleoid/plasmid/game/map/template/TemplateRegion.java
```diff
@@ -20,14 +22,26 @@ public BlockBounds getBounds() {
...
         this.bounds.serialize(tag);
+        tag.put("data", this.data);
         return tag;
     }
 
     public static TemplateRegion deserialize(CompoundTag tag) {
         String marker = tag.getString("marker");
-        return new TemplateRegion(marker, BlockBounds.deserialize(tag));
+        CompoundTag data;
+        if (!tag.contains("data")) { data = new CompoundTag(); } else { data = tag.getCompound("data"); }
```
gegy1000, line 44: I think `getCompound()` should return an empty tag object if it's not present

gegy1000:
Oh, another thing to consider is how the data should be passed to the map regions in the compiled MapTemplate, since this is currently just within the staging template.

LambdAurora:
Wouldn't it be better to pass the TemplateRegion instead of just BlockBounds for all the region calls in the compiled one?

gegy1000:
We'll definitely need to expose that somehow. Ideally we can do that without causing any breaking changes, but if we do need to change the API it's probably not too bad.

src/main/java/xyz/nucleoid/plasmid/command/MapCommand.java
```diff
@@ -268,4 +318,34 @@ private static int commitRegion(CommandContext<ServerCommandSource> context) thr
             );
         };
     }
+
+    @FunctionalInterface
+    private interface MapExecutor {
+        int execute(CommandContext<ServerCommandSource> context, StagingMapTemplate map, BlockPos pos) throws CommandSyntaxException;
+
+        static Command<ServerCommandSource> make(MapExecutor executor) {
```
gegy1000, line 326: It might be simpler to have this work as like `StagingMapManager getMap(CommandContext<ServerCommandSource>) throws CommandSyntaxException`?

gegy1000 (approved):

PR #51, opened 2020-11-18
Fix projectile events not passing in the projectile

SuperCoder7979:
This is a breaking change and should be merged in the next version.


Gegy:
Merged into staging branch in https://github.com/NucleoidMC/plasmid/commit/eaed315d004a007b2d85bd54bf53c383f69df63c

PR #56, opened 2020-11-19
Refactor drop item game rule into event

SuperCoder7979:
This allows minigames to give players the ability to drop specific items. Reviews welcome.

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/game/GameWaitingLobby.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/GameWaitingLobby.java
@@ -47,22 +47,18 @@ public final class GameWaitingLobby {
             game.setRule(GameRule.FALL_DAMAGE, RuleResult.DENY);
             game.setRule(GameRule.HUNGER, RuleResult.DENY);
             game.setRule(GameRule.INTERACTION, RuleResult.DENY);
+            game.setRule(GameRule.THROW_ITEMS, RuleResult.DENY);
 
             game.on(GameTickListener.EVENT, lobby::onTick);
             game.on(RequestStartListener.EVENT, lobby::requestStart);
             game.on(OfferPlayerListener.EVENT, lobby::offerPlayer);
             game.on(PlayerRemoveListener.EVENT, lobby::onRemovePlayer);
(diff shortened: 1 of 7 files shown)


SuperCoder7979:
Having both the event and the rule raises some questions, i.e. when would the rule be used and when would the event be used. Maybe we can do it where the rule takes precedence over the event, so the event only fires if the rule is true?

Gegy (approved):

LambdAurora (approved):

Restioson:
That sounds like a footgun. Considering how it's only a one line change if you were using the rule, is it worth it?

SuperCoder7979:
The use case for it here is that I want people to share certain things in minigames, while not sharing others. For example, in Wave Defense I want players to share iron and food while not being able to drop their weapons and armor, and this provides a nice utility to do so. As for the error, I'm hoping that minigame devs read the javadoc- but I suppose it makes sense that the item drop listener wouldn't work if item drops are disabled :P

Restioson:
Yea I get the use case for the listener, but it seems to be extremely simple to implement the rule in terms of the listener (just return fail), which is a one liner. So couldn't existing users of the rule just use that and then the rule is removed in favour of the event?

SuperCoder7979:
Hmm, I'm not sure... I assumed we were keeping the rule for backwards compat, but I guess we'll need to decide on that at some point.

LambdAurora:
Well, 0.4 is already breaking changes so keeping backward compat is not the biggest worry :P

I do agree on the duplicate nature of the rule, would be better to only have the event I guess.

Gegy:
All rules that we have could be implemented with an event in some form. If we’re choosing events over rules, we may as well remove all rules?

But- the point of this rules system is to eventually extract into its own library similar to WorldGuard, where we could use these rules to protect the spawn, for example.

Restioson:
> . If we’re choosing events over rules, we may as well remove all rules?

Restioson:
FUCK

Gegy:
Merged into staging branch in https://github.com/NucleoidMC/plasmid/commit/c9dfd54c88e0fd21b959e9ecce5f2de3841aa290

PR #55, opened 2020-11-19
Fully separate bubble world time and weather & never inherit gamerules

Gegy:

PR #52, opened 2020-11-19
Refactor GameWorld and player management

Gegy:
This PR involves some general refactoring relating to how GameWorlds are managed, as well as how players are managed.
- Rename `GameWorld` -> `GameSpace`
- Split `GameSpace` into `GameSpace` interface and `ManagedGameSpace` implementation. This disallows game implementations from directly interacting with parts of `GameSpace` logic that they shouldn't be, such as adding/removing players or invoking events.
- Refactor `PlayerSet` into hierarchy: `MutablePlayerSet` -> `PlayerSet` -> `PlayerOps` to allow for more specific usage
- Remove listener system from `PlayerSet`: poor design of ownership and state management
- `PlayerSet` now stores `UUID`s instead of `ServerPlayerEntity` instances. Storing entity instances can cause issues, especially given a player can leave and rejoin.
- Add `GlobalWidgets`: a utility for dealing with widgets now that `PlayerSet` listeners are removed.
- Refactor game initializers: now return a `GameOpenProcedure` and runs off-thread. No more CompletableFutures!

PR #58, opened 2020-11-20
Extract bubble worlds into library & add map workspace dimensions

Gegy:
This is rebased on top of #55 and #52 since this works over both of those branches, so those will need to be merged first.


LambdAurora:
noice

PR #59, opened 2020-11-21
Make Scoreboards Translatable (Updated)

arthurbambou:


LambdAurora (approved):

Gegy (changes requested):
I think unfortunately by the nature of `Text` and scoreboards, exposing access to send Text over the scoreboard is misleading within the API, given anything but basic translation/literal will not be rendered. We'll need to expose the ability to write translated messages somehow else.

arthurbambou:
@Gegy Can you review again?
I changed it so it doesn’t use a Text anymore

arthurbambou:
Close in favor of #63

PR #61, opened 2020-11-22
Networking documentation/specification

LambdAurora:
Basically, the documentation/specification of the networking protocol.

What must be thought in the design:
 - We must not exceed the 1MB limit of packets
 - All of the features allowed by this networking MUST not give an advantage to the players having a client that implements the protocol.
   - The only exception for this rules is the tools to create/build maps and moderation.

This is still a draft, despite it being a draft reviews are very welcomed and needed to build a good networking protocol for plasmid :)

### DIFF
--- a/doc/network.md
+++ b/doc/network.md
@@ -1,0 +1,124 @@
+# Plasmid Networking
+
+Plasmid networking uses the custom payload packets of Minecraft. This document describes the networking with Plasmid.
+
+## Specification
+
+### Packet Specification
+
+In this specification a packet is its own section, 
+a description of what the packet is used for and how it should be handled,
+then it specifies the direction (`C->S`, `C<-S`, `C<->S`),
+and a table of the fields and type of the packet.
+
+## Types
+
+### BlockPos
+
+`BlockPos` type represents a block position in the world,
+`x` as a 26-bit integer, followed by `y` as a 12-bit integer, followed by `z` as a 26-bit integer (all signed, two's complement).
+Currently, encoded as a 64-bit integer split into three parts:
+ - x: 26 MSBs
+ - z: 26 middle bits
+ - y: 12 LSBs
+
+### Bounds
+
+Bounds represents a box which are defined with two [BlockPos]: a minimum and a maximum.
+
+| Fields | Type       | Description             |
+|:------:|:----------:|:------------------------|
+| min    | [BlockPos] | The minimum coordinate. |
+| max    | [BlockPos] | The maximum coordinate. |
+
+### Region
+
+A region represents a named box in a Plasmid map which can contain arbitrary data.
+
+| Fields | Type       | Description                          |
+|:------:|:----------:|:-------------------------------------|
+| marker | string     | The name of the region.              |
+| bounds | [Bounds]   | The bounds of the region.            |
+| data   | NBT Tag    | Arbitrary data stored in the region. |
+
+## Packets
+
+### `plasmid:staging_map/enter`
+
+Packet sent when a player enters a staging map editor. The client should render the bounds of the map if present,
+and the contained regions, points, etc.  
+A `plasmid:staging_map/regions` packet should be sent with this packet if the staging map contains regions.
+
+Direction: `C<-S`
+
+| Fields |  Type      | Description                                                                    |
+|:------:|:----------:|:-------------------------------------------------------------------------------|
+| id     | Identifier | The identifier of the entered staging map.                                     |
+| bounds | [Bounds]   | The bounds of the staging map, may be `[[0, 0, 0], [0, 0, 0]]` if not present. |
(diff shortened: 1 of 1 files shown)


doc/network.md
```diff
@@ -0,0 +1,325 @@
...
+#### `plasmid:workspace/data`
+
+Sets the associated data on the workspace map.
+A client can request data change, but can be rejected. If a request is accepted, the packet is sent back to the client.
+
+Direction: `C<->S`
+
+| Fields | Type       | Description                               |
+|:------:|:----------:|:------------------------------------------|
+| id     | Identifier | The identifier of the affected workspace. |
```
Gegy, line 142: This is sometimes referred to as just `id` and sometimes as `workspace`- maybe we should standardise them all as `workspace`? Or `workspace_id` if we need to be more specific

doc/network.md
```diff
@@ -0,0 +1,325 @@
...
+
+Specifies a single region, the client should render a box, and a marker tag for it.
+If sent from a client, this is considered as an update request which can be rejected.
+
+Direction: `C<->S`
+
+| Fields    | Type       | Description                                                          |
+|:---------:|:----------:|:---------------------------------------------------------------------|
+| workspace | Identifier | The identifier of the workspace in which the region should be added. |
+| id        | [VarI32]   | The region runtime identifier.                                       |
```
Gegy, line 155: Might be worth also qualifying these as `runtime_id` or `network_id` specifically

doc/network.md
```diff
@@ -0,0 +1,325 @@
...
+
+Specifies regions to the client, for each region the client should render a box, and a marker tag.
+
+Direction: `C<-S`
+
+| Fields    | Type                              | Description                                                       |
+|:---------:|:---------------------------------:|:------------------------------------------------------------------|
+| workspace | Identifier                        | The identifier of the workspace containing the following regions. |
+| marker    | string                            | The marker of the regions to update.                              |
+| regions   | ([VarI32], [Bounds], NBT Tag)\[\] | The array of regions to update. The tuple is, in order, `(runtime id, bounds, data)`. |
```
Gegy, line 168: If a lot of regions with different markers were to be updated, would they be sent in separate packets?

doc/network.md
```diff
@@ -0,0 +1,325 @@
...
+Specifies a single region, the client should render a box, and a marker tag for it.
+If sent from a client, this is considered as an update request which can be rejected.
+
+Direction: `C<->S`
+
+| Fields    | Type       | Description                                                          |
+|:---------:|:----------:|:---------------------------------------------------------------------|
+| workspace | Identifier | The identifier of the workspace in which the region should be added. |
+| id        | [VarI32]   | The region runtime identifier.                                       |
+| region    | [Region]   | The region to add.                                                   |
```
Gegy, line 156:
```suggestion
| region    | [Region]   | The region to add or update.                                         |
```

doc/network.md
LambdAurora: Yes, and even if they weren't, there's still the 1MB limit so I guess there's not much choice.

Restioson:
Does this require anything else? It would be nice to use this to make a client mod for setting up regions for Dungeons.

LambdAurora:
There's definitely missing stuff, but meh, the current stuff here is "good enough".

doc/network.md
```diff
@@ -0,0 +1,246 @@
...
+
+Direction: `C->S`
+
+| Fields    | Type           | Description                                                            |
+|:---------:|:--------------:|:-----------------------------------------------------------------------|
+| workspace | Identifier     | The identifier of the new workspace.                                   |
+| bounds    | [Bounds]       | The bounds of the map, may be `[[0, 0, 0], [0, 0, 0]]` if not present. |
+| generator | string (32767) | The generator string for the map, may be empty if unspecified.         |
+| has_data  | bool           | `true` if contains a data NBT tag, else `false`.                       |
+| data      | (NBT Tag)?     | Arbitrary data assigned to the map template.                           |
```
Gegy, line 124: The optional data tag like this doesn't seem to be used anywhere else- may be worth just forcing to send an empty data tag instead of sending none.

doc/network.md
```diff
@@ -0,0 +1,246 @@
...
+#### `plasmid:workspace/region`
+
+Specifies a single region, the client should render a box, and a marker tag for it.
+If sent from a client, this is considered as an update request which can be rejected.
+
+Direction: `C<->S`
+
+| Fields     | Type       | Description                                                          |
+|:----------:|:----------:|:---------------------------------------------------------------------|
+| workspace  | Identifier | The identifier of the workspace in which the region should be added. |
```
Gegy, line 169: With all these packets we can probably exclude the workspace identifier and force the assumption that it is the current workspace. This helps with network traffic as well as implicit security in the C->S direction.

Gegy (approved):

PR #67, opened 2020-11-25
Add a rule for sanitizing creative actions

haykam821:
This pull request adds the `SANITIZE_CREATIVE_ACTIONS` rule. Unless explicitly disabled, this removes all NBT tags from an  item stack picked in creative except the following:

- `Damage`
- `Unbreakable`
- `BucketVariantTag` used for tropical fish buckets
- `color` used for leather armor
- `SkullOwner` used for player heads

This rule breaks the use of valid potions and enchantment books, and clears all items from shulker boxes. Special cases will need to be made for these scenarios.

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/game/rule/GameRule.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/rule/GameRule.java
@@ -12,4 +12,5 @@ public final class GameRule {
     public static final GameRule THROW_ITEMS = new GameRule();
     public static final GameRule UNSTABLE_TNT = new GameRule();
     public static final GameRule TEAM_CHAT = new GameRule();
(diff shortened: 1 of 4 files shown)


Snorps:
:(

Gegy (approved):

LambdAurora (commented):
Good, but there's more NBT I would like to see whitelisted:

 - `map` for filled map items
 - `LodestoneTracked`, `LodestonePos`, `LodestoneDimension` for compass

We can whitelist `Items` for the 1.17 bundle item, but that would require to sanitize all the listed items.
`BlockEntityTag` could be whitelisted but only under certain conditions I guess.

Gegy:
Would we not want to apply restrictions such that users can only access what they would be able to with normal creative mode?

Patbox:
Closing this as it's related to old version of plasmid. If you want to remake it using newer one, feel free to create new pr

PR #72, opened 2020-11-27
Add listener for a player firing an arrow

Restioson:
Crossbow support is not there quite yet, but I've tested bow support in koth and cancellation works.

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/game/event/PlayerFireArrowListener.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/event/PlayerFireArrowListener.java
@@ -23,7 +23,6 @@ public interface PlayerFireArrowListener {
             ActionResult result = listener.onFireArrow(user, tool, arrows, remaining, projectile);
 
             if (result != ActionResult.PASS) {
-                System.out.println("Returning fail");
                 return result;
             }
         }
--- a/src/main/java/xyz/nucleoid/plasmid/game/event/PlayerFireArrowListener.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/event/PlayerFireArrowListener.java
@@ -1,0 +1,35 @@
+package xyz.nucleoid.plasmid.game.event;
+
+import net.minecraft.entity.projectile.ArrowEntity;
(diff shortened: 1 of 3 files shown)


Restioson:
It's more difficult to write a mixin for crossbow since it uses static methods to dispatch this, and they're quite broken up, so I'm not sure where to inject the code. For example, it calls `shootAll` which has the shooting logic and then `setCharged` afterward. Ideally returning `ActionResult.FAIL` would not reset the charge on the crossbow, so it would need to be mixed in to before the `shootAll` call. However, by this time none of the arrow entities have been created, so no useful information can be passed forward.

src/main/java/xyz/nucleoid/plasmid/game/event/PlayerFireArrowListener.java
```diff
@@ -0,0 +1,35 @@
...
+ * <li>{@link ActionResult#FAIL} cancels further handlers and does not execute vanilla behavior.
+ * <li>{@link ActionResult#PASS} moves on to the next listener.</ul>
+ */
+public interface PlayerFireArrowListener {
+    EventType<PlayerFireArrowListener> EVENT = EventType.create(PlayerFireArrowListener.class, listeners -> (user, tool, arrows, remaining, projectile) -> {
+        for (PlayerFireArrowListener listener : listeners) {
+            ActionResult result = listener.onFireArrow(user, tool, arrows, remaining, projectile);
+
+            if (result != ActionResult.PASS) {
+                System.out.println("Returning fail");
```
Gegy, line 26: Debug print :P

Restioson: Oof I thought this printed from test minigame 😳

Gegy:
I think it's fair to merge this PR without support for crossbows, since the event signature shouldn't have to change later if we were to add support for them.

Restioson:
> I think it's fair to merge this PR without support for crossbows, since the event signature shouldn't have to change later if we were to add support for them.

Alright, I'll mark it ready for review and then we can open an issue for crossbows.

Issue #68, opened 2020-11-25
Faked content is not persistent


Gegy:

Gegy:
This is invalid and rather a result of faked content not being registered.

PR #100, opened 2021-01-07
Clean up fakes

ramidzkh:
Quick cleanup while I was looking at #68


src/main/java/xyz/nucleoid/plasmid/fake/FakeItem.java
```diff
@@ -16,26 +15,20 @@
...
+     * Resolves the proxy {@link ItemStack} instance to be sent to the client. If the associated {@link Item} of this
+     * {@link ItemStack} is an instance of {@link FakeItem}, the {@link #asProxy(ItemStack)} method will be used
+     *
+     * <p>This method is intended for use in case the item has information which the client would not be able to
+     * normally see, so that lore could be appended, or its visuals changed
+     *
+     * @param stack The server side {@link ItemStack}
+     * @return The {@link ItemStack} to send to the client
+     */
     static ItemStack getProxy(ItemStack stack) {
```
Gegy, line 32: I believe there were issues when proxying ItemStacks with Server-Translations- is this tested?

ramidzkh: They don't seem to work; I'll try to fix this now

src/main/java/xyz/nucleoid/plasmid/mixin/fake/PacketByteBufMixin.java
```diff
@@ -18,6 +24,6 @@
             )
     )
     private Item modifyItem(ItemStack stack) {
-        return FakeItem.getProxy(stack.getItem());
+        return Fake.getProxy(stack.getItem());
```
Gegy, line 27: This is now essentially duplicated with the `modifyItemStack` mixin

ramidzkh: `modifyItemStack` doesn't have to change the item type. Maybe it should do that actually 🤔

Gegy: Oh- definitely, lol.

src/main/java/xyz/nucleoid/plasmid/fake/FakeItem.java
Gegy: I believe the reason for this not working is that Server-Translations depends on the ItemStack's Item being the source fake item when it checks to translate the name. 
https://github.com/arthurbambou/Server-Translations/blob/abc6e0c4594fa2a8bbb195cfff7062678da5a86c/api/src/main/java/fr/catcore/server/translations/api/mixin/packet/PacketByteBufMixin.java#L50-L57

Gegy:
The fake API has since been replaced with Polymer

PR #101, opened 2021-01-07
Use game IDs where possible for messages

Geek202:
Replaces `ConfiguredGame.getName()` with two new methods `getOptionalName()` and `getDisplayName(gameId)` that fix the broken behavior shown below:
image
(use of the `GameType` id rather than the id of the game config itself)


Gegy (approved):

diabolical17:
Rather than removing `ConfiguredGame.getName()`

Mark it as deprecated.

Gegy:
@diabolical17 it is not removed, it is deprecated.

diabolical17:
Oops my bad, mobile GitHub said it was deleted.

Issue #104, opened 2021-01-12
Split events into scoped events library


Gegy:
This will be useful for leukocyte, but also for many other cases where events are useful without having a Plasmid minigame running.

This could potentially include rules too which we might be able to couple more closely with the related events, since a common pattern is for a rule to match an event exactly. In this case, leukocyte would become just a wrapper around this library that supports setting these rules with persistence from a command interface.

Restioson:
IMO, only the innermost (most nested in the scope) event handler should run. Otherwise, if they all ran, you would only be able to add new event handlers which cancel the event but not explicitly allow it, because cancellation always takes precedence over allowing.

Gegy:
This is now resolved in 0.5 with the integration of [Stimuli](https://github.com/NucleoidMC/stimuli).

PR #107, opened 2021-01-13
Added PLAYER_PROJECTILE_KNOCKBACK rule

PotatoPresident:


Restioson:
I believe the plan is to switch to using leukocyte directly, and as such most/all(?) rules should be implemented there. Not sure if that has changed or applies to all rules but see [this comment](https://github.com/NucleoidMC/plasmid/blob/ae1a56b6948ea316fd32575968652b3e19dec86d/src/main/java/xyz/nucleoid/plasmid/game/rule/GameRule.java#L6)

Gegy:
This is fine for now. Given we plan to extract rules and events further into their own library (see #104), there'll need to be a lot of shifting around anyway. For now, this is so specific that I'm not sure it would even fit into leukocyte.

Gegy (approved):

Issue #105, opened 2021-01-12
Statistics API


Gegy:
Plasmid should have an API for minigames to provide statistics for that given game. The use of this data can be up to external mods- in our case, nucleoid-extras, which will send these to the backend.

The API needs to be designed in such a way that each data point can indicate to the backend how it should be stored & displayed. For example, some may want to be stored as an average while others would be as a total.

Geek202:
An idea for getting the statistics data out of plasmid and into other mods (nucleoid-extras) could be to add some Fabric API style events to plasmid for games being started/ending that provide access to data about the game such as the statistics. These events could also be used by other mods to show information when games are opened, for example to send statistics about a game to the client.

Gegy:
Ah, yep! Events will definitely be needed to be exposed. :)

Geek202:
I can take a look at seeing if I can implement a global events system that can be expanded as this gets implemented if you like. I'd assume the main ones would be game start/end, and potentially also some related to players joining/leaving games too.

Gegy:
That would be wonderful! Thanks

PR #108, opened 2021-01-14
Start work on events for use by other mods

Geek202:
This is currently a work in progress.
TODO:
- Game events:
  - [x] Game opening
  - [x] Game opened
  - [x] Game start requested
  - [x] Game closing
  - [x] Game logic set (player listeners can be registered with existing apis on the `GameLogic`)
- ~~Player events:~~
  - [ ] ~~Player join game~~
  - [ ] ~~Player leave game~~

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/Plasmid.java
+++ b/src/main/java/xyz/nucleoid/plasmid/Plasmid.java
@@ -57,12 +57,6 @@ public final class Plasmid implements ModInitializer {
         GameType.register(new Identifier(Plasmid.ID, "random"), RandomGame::open, RandomGameConfig.CODEC);
 
         this.registerCallbacks();
-
-        // TODO: Remove. This is for testing.
-//        GameEvents.OPENING.register(((gameId, game, oneShot) -> LOGGER.info("GAME EVENTS DEBUG | id: {}, game: {}, oneShot: {}", gameId, game, oneShot)));
-//        GameEvents.OPENED.register(((gameId, game, oneShot, gameSpace) -> LOGGER.info("GAME EVENTS DEBUG | id: {}, game: {}, oneShot: {}, gameSpace: {}", gameId, game, oneShot, gameSpace)));
-//        GameEvents.START_REQUEST.register((gameSpace, result) -> LOGGER.info("GAME EVENTS DEBUG | gameSpace: {}, result: {}", gameSpace, result));
-//        GameEvents.CLOSING.register(((gameSpace, reason) -> LOGGER.info("GAME EVENTS DEBUG | gameSpace: {}, reason: {}", gameSpace, reason)));
     }
 
     private void registerCallbacks() {
--- a/src/main/java/xyz/nucleoid/plasmid/Plasmid.java
+++ b/src/main/java/xyz/nucleoid/plasmid/Plasmid.java
@@ -60,7 +60,7 @@ public final class Plasmid implements ModInitializer {
 
         // TODO: Remove. This is for testing.
         GameEvents.OPENING.register(((gameId, game, oneShot) -> LOGGER.info("GAME EVENTS DEBUG | id: {}, game: {}, oneShot: {}", gameId, game, oneShot)));
-        GameEvents.OPEN.register(((gameId, game, oneShot, gameSpace) -> LOGGER.info("GAME EVENTS DEBUG | id: {}, game: {}, oneShot: {}, gameSpace: {}", gameId, game, oneShot, gameSpace)));
+        GameEvents.OPENED.register(((gameId, game, oneShot, gameSpace) -> LOGGER.info("GAME EVENTS DEBUG | id: {}, game: {}, oneShot: {}, gameSpace: {}", gameId, game, oneShot, gameSpace)));
         GameEvents.START_REQUEST.register((gameSpace, result) -> LOGGER.info("GAME EVENTS DEBUG | gameSpace: {}, result: {}", gameSpace, result));
         GameEvents.CLOSING.register(((gameSpace, reason) -> LOGGER.info("GAME EVENTS DEBUG | gameSpace: {}, reason: {}", gameSpace, reason)));
     }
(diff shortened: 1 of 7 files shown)


Geek202:
Should also mention this related to #105

src/main/java/xyz/nucleoid/plasmid/game/ManagedGameSpace.java
```diff
@@ -270,7 +271,9 @@ private void onRemovePlayer(ServerPlayerEntity player) {
     public CompletableFuture<StartResult> requestStart() {
         return Scheduler.INSTANCE.submit(server -> {
             try {
-                return this.invoker(RequestStartListener.EVENT).requestStart();
+                StartResult startResult = this.invoker(RequestStartListener.EVENT).requestStart();
+                GameEvents.START_REQUEST.invoker().onRequestStart(this, startResult);
```
Gegy, line 275: Should these events have the ability to replace the start result? It might be worth also noting in the documentation that they can be called with a failed start result.

src/main/java/xyz/nucleoid/plasmid/event/GameEvents.java
```diff
@@ -0,0 +1,72 @@
...
+         * @param gameId The ID of the game config being opened
+         * @param game The game and its configuration
+         * @param oneShot If the game is a one off, or whether it is part of a channel
+         */
+        void onGameOpening(Identifier gameId, ConfiguredGame<?> game, boolean oneShot);
+    }
+
+    public interface GameOpen {
+        /**
+         * @param gameId THe ID of the game config that has been opened
```
Gegy, line 57:
```suggestion
         * @param gameId The ID of the game config that has been opened
```

src/main/java/xyz/nucleoid/plasmid/event/GameEvents.java
```diff
@@ -0,0 +1,72 @@
...
+    private GameEvents() { }
+
+    public static final Event<GameOpening> OPENING = EventFactory.createArrayBacked(GameOpening.class,
+            listeners -> (gameId, game, oneShot) -> {
+        for (GameOpening listener : listeners) {
+            listener.onGameOpening(gameId, game, oneShot);
+        }
+    });
+
+    public static final Event<GameOpen> OPEN = EventFactory.createArrayBacked(GameOpen.class,
```
Gegy, line 25: Might be worth clarifying this as `OPENED`, similar to how server starting/started events are in Fabric API.

src/main/java/xyz/nucleoid/plasmid/event/GameEvents.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public interface GameOpen {
+        /**
+         * @param gameId THe ID of the game config that has been opened
+         * @param game The game and its configuration
+         * @param oneShot true if the game is a one off, false if it is part of a channel
+         * @param gameSpace The {@link GameSpace} the game is running in.
+         */
+        void onGameOpen(Identifier gameId, ConfiguredGame<?> game, boolean oneShot, GameSpace gameSpace);
```
Gegy, line 62: A bit conflicted whether it makes sense for this event to be called from the responsible channels as it is now, or rather whether it should be called from deeper inside the game opening logic itself which they both go through.
Also, is the `oneShot` parameter needed?

Geek202: I call the events from in the channels because it gives more access to context (game ID mostly) than inside `ConfiguredGame#open`. The `oneShot` parameter is there to give as much context as possible; for example, a mod may not what to do something when games are opened from a command

src/main/java/xyz/nucleoid/plasmid/game/ManagedGameSpace.java
Geek202: Being able to override the start result could be useful

src/main/java/xyz/nucleoid/plasmid/event/GameEvents.java
Gegy: Hmm- what about having like a `OneshotChannelOpen` event and then the regular game opening events which don't have access to game ID, but will always be called?

Generally relying on game config IDs is something to avoid, given they are not mandatory and it is valid to create an anonymous game config.

Geek202: Ah, I didn't fully realise it was possible to open games anonymously, I'll see if I can add a separate event for oneshots then

src/main/java/xyz/nucleoid/plasmid/game/ManagedGameSpace.java
```diff
@@ -270,7 +272,9 @@ private void onRemovePlayer(ServerPlayerEntity player) {
...
```
Gegy, line 276: The result from this looks like it is never used?

Geek202: Fixed!

src/main/java/xyz/nucleoid/plasmid/game/ManagedGameSpace.java
```diff
@@ -338,6 +341,9 @@ public void close(GameCloseReason reason) {
 
         Scheduler.INSTANCE.submit(server -> {
             try {
+                // TODO: Hook game closing event here
+                GameEvents.CLOSING.invoker().onGameClosing(this, reason);
```
Gegy, line 345: Is this TODO still relevant? This can also be renamed to `onGameSpaceClosing` for consistency

src/main/java/xyz/nucleoid/plasmid/Plasmid.java
```diff
@@ -56,6 +57,12 @@ public void onInitialize() {
         GameType.register(new Identifier(Plasmid.ID, "random"), RandomGame::open, RandomGameConfig.CODEC);
 
         this.registerCallbacks();
+
+        // TODO: Remove. This is for testing.
+//        GameEvents.OPENING.register(((gameId, game, oneShot) -> LOGGER.info("GAME EVENTS DEBUG | id: {}, game: {}, oneShot: {}", gameId, game, oneShot)));
```
Gegy, line 62: These are probably ready to remove :)

Gegy (approved):
Perfect, thank you so much!

PR #114, opened 2021-02-22
Add the locate subcommand to the game command

### LINKED ISSUE #79: Add a /game locate subcommand
It would be useful for players to be able to instantly find which minigame another player is using a command such as `/game locate <username>`. If the given player is in a game, a link to join that game should also be available.

haykam821:
This pull request adds the `locate` subcommand. It can be used to find which channel a player is in, and sends a link to join the channel.

Fixes #79

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/game/config/GameConfigs.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/config/GameConfigs.java
@@ -80,6 +80,11 @@ public final class GameConfigs {
         return CONFIGURED_GAMES.get(identifier);
     }
 
+    @Nullable
+    public static Identifier getIdentifier(ConfiguredGame<?> game) {
+        return CONFIGURED_GAMES.getIdentifier(game);
+    }
+
     public static Set<Identifier> getKeys() {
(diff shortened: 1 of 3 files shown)


src/main/java/xyz/nucleoid/plasmid/game/config/GameConfigs.java
```diff
@@ -80,6 +80,11 @@ private static Identifier identifierFromPath(Identifier location) {
         return CONFIGURED_GAMES.get(identifier);
     }
 
+    @Nullable
+    public static Identifier getIdentifier(ConfiguredGame<?> game) {
```
Gegy, line 84: This shouldn't be provided in the API, given not all games have identifiers and we are currently trying to move away from using them as much as possible.

Gegy (changes requested):

src/main/java/xyz/nucleoid/plasmid/game/config/GameConfigs.java
haykam821: Oops, this was unintentional.

Gegy (approved):

PR #125, opened 2021-02-24
Pass the reason to the game close listener

haykam821:
This pull request makes the `GameCloseListener` interface have a new parameter for a `GameCloseReason`. This is passed normally, except for when a game closes due to a game opening, where it will be passed `null`.


src/main/java/xyz/nucleoid/plasmid/game/ManagedGameSpace.java
```diff
@@ -174,7 +174,7 @@ public void openGame(Consumer<GameLogic> builder) {
             closedGameLogic.getResources().close();
 
             try {
-                closedGameLogic.getListeners().invoker(GameCloseListener.EVENT).onClose();
+                closedGameLogic.getListeners().invoker(GameCloseListener.EVENT).onClose(null);
```
Gegy, line 177: Maybe this should have a specific close reason type for switching?

haykam821: Fixed in 1ad427eb96fbed995728f81c7e712224efe0f68c

PR #126, opened 2021-02-25
Add game packets

haykam821:
This pull request adds a simple set of custom payload packets that are sent to clients when a player is added or removed from a game space, or a game space is closed.

### DIFF
--- a/doc/network.md
+++ b/doc/network.md
@@ -106,6 +106,21 @@ Direction: `C<-S`
 | player_count   | int        | The new player count of the game space after the player joined. |
 | player_uuid    | UUID       | The UUID of the player that joined the game space.              |
 
+#### `plasmid:game/player_removed`
+
+Packet sent when a player is remove from a game space.
+
+Direction: `C<-S`
+
+| Fields         | Type       | Description                                                         |
+|:--------------:|:----------:|:--------------------------------------------------------------------|
+| game_type_id   | Identifier | The identifier of the game type.                                    |
+| game_type_name | Text       | The name of the game type.                                          |
+| game_id        | Identifier | The identifier of the game.                                         |
+| game_name      | Text       | The name of the game.                                               |
+| player_count   | int        | The new player count of the game space after the player is removed. |
+| player_uuid    | UUID       | The UUID of the player that was removed from the game space.        |
+
 ### Workspace-related Packets
 
 The server controls most of the requests sent by a client for workspace-related packets as it requires some permissions.
--- a/doc/network.md
+++ b/doc/network.md
@@ -93,7 +93,8 @@ Size: 0 if non-present, else size of `X`.
 
 #### `plasmid:game/player_add`
 
-Packet sent when a player is joined to a game space.
+Packet sent to all players of a game space when a player is joined to a game space.
+The player that was joined to the game space also receives this packet.
 
 Direction: `C<-S`
 
@@ -108,7 +109,8 @@ Direction: `C<-S`
 
 #### `plasmid:game/player_removed`
 
-Packet sent when a player is remove from a game space.
+Packet sent to all players of a game space when a player is removed from a game space.
+The player that was removed from the game space also receives this packet.
 
 Direction: `C<-S`
 
@@ -123,7 +125,7 @@ Direction: `C<-S`
 
 #### `plasmid:game/game_close`
 
-Packet sent when a game space closes.
+Packet sent to all players in a game space when it closes.
 
 Direction: `C<-S`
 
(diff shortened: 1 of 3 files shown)


Gegy (commented):
I worry that this could get quite spammy on the network if there are a lot of players, where the information may not be used.

haykam821:
Would a way to opt into game packets make sense?

Gegy (approved):

doc/network.md
```diff
@@ -89,6 +89,54 @@ Size: 0 if non-present, else size of `X`.
...
+Packet sent to all players of a game space when a player is joined to a game space.
+The player that was joined to the game space also receives this packet.
+
+Direction: `C<-S`
+
+| Fields         | Type       | Description                                                     |
+|:--------------:|:----------:|:----------------------------------------------------------------|
+| game_type_id   | Identifier | The identifier of the game type.                                |
+| game_type_name | Text       | The name of the game type.                                      |
+| game_id        | Identifier | The identifier of the game.                                     |
```
Gegy, line 105: On second thought, I'm not sure if this should be included in the packet. Relying on game ids is never a good idea.

haykam821: Is there any reason why packet consumers should not be able to have access to the game ID?

Gegy: They can’t and shouldn’t be relied on by anything. Games don’t have an ID- hence the rename to source. It’s just where the game was loaded from, and that might not even be present in the case of an anonymously created game.

haykam821: Packet consumers would be required to handle anonymous games, then. If the identifier was removed, packet consumers would just end up relying on text, which is worse. After all, games can have no text as well (though, this just falls back rather than being `null`).

Gegy: They should probably rather be passed all the game custom values.

haykam821: Most games don't pass any custom values, and if they did it would be at the mercy of the game's developers, similar to the name.

Gegy: Hardcoding per JSON file is arguably worse. It can change, and new JSONs can be added and then the handling needs to be updated. Nothing should rely on the JSON file that creates the config- the config itself must provide enough information in itself. I think it's fine if the consumer must rely on the game providing the information it needs.

haykam821: Packet consumers won't be stopped from hardcoding things; it'll just make packets harder to consume. I think the best solution is to document that it is generally a bad idea to depend on game IDs for longer-term cases.

Gegy: Giving the option to use game sources will encourage it, where by design consumer should instead make use of custom data values to be provided by games.

haykam821: Custom data values would be useless or non-existent in most cases. I see the game ID as the only solution to certain use cases, where anonymous games can be handled as well just fine.

Gegy: Can you elaborate on these cases?

haykam821: For example, a tracker client mod would store the games in a map like the following:

```json
{
    "bedwars:the_end_bed_wars": 1,
    "bedwars:cubes_bed_wars": 1
}
```

Anonymous games would never appear in this map due to a lack of a game ID.

Gegy: Why would it store things like this as opposed to having some value in the game's config?

haykam821: Would that custom value not just be the same as the game ID? In addition, would all anonymous games have that custom value as well? It would duplicate the existing game ID for not much benefit.

Gegy: No- it would provide whatever values are needed directly. Anonymous games are an example of where things would not have a game id, but really the motivation in this case is they can't be considered to be stable. A new game config can be added, or a config can be moved around- but then you have to update the map within the tracker to consider that.

haykam821: The game ID is the thing that this map would need, and for the problem the tracker aims to solve, there is no way around it; any attempts at providing a custom value replacement would just duplicate the ID.

I think packet consumers can handle the task of recognizing that the given game IDs are not stable and may not exist. The tracker would ignore games without IDs and treat games with changed IDs as entirely separate games, unless the user chooses to merge two game IDs.

Gegy: The point is to not have the map in the first place, but rather the values within the map provided within the game. What do you imagine this map would be used to store?

haykam821: The map would store play counts for each game.

LambdAurora (changes requested):
`plasmid:game/player_add` needs to be split:
`plasmid:game/self_add` would contain the same fields as the current packet but without `player_uuid`
The new `plasmid:game/player_add` would be trimmed down (be gone all the fields starting with `game_`)

`plasmid:game/player_remove` would only contain `player_count` and `player_uuid` (all data would be cached by the clients when joining a game anyway)

same with `plasmid:game/close` (the `game_` can be dropped here)
only the reason field is needed.

PR #127, opened 2021-02-25
Pass the explosion instead of its affected blocks to the explosion listener

haykam821:


Gegy (approved):

src/main/java/xyz/nucleoid/plasmid/mixin/game/event/ExplosionMixin.java
```diff
@@ -30,7 +30,7 @@ private void affectWorld(boolean blocks, CallbackInfo ci) {
         ManagedGameSpace gameSpace = ManagedGameSpace.forWorld(this.world);
         if (gameSpace != null) {
             try {
-                gameSpace.invoker(ExplosionListener.EVENT).onExplosion(this.affectedBlocks);
+                gameSpace.invoker(ExplosionListener.EVENT).onExplosion(this);
```
Gegy, line 33: On second thought, is `affectedBlocks` exposed and accessible to users?

haykam821: Yes, there is a public getter method available. I realized, though, that the `blocks` (actually `particles`) parameter is not passed and the type is incorrect (`ExplosionMixin` rather than `Explosion`).

PR #133, opened 2021-04-06
MODIFY_INVENTORY and MODIFY_ARMOR gamerule

Geek202:
Adds a new gamerule for minigames that specifies whether players can move items around in their inventories. The villager 'no' sound is also played if users can't move items, however there are some cases where it doesn't (mostly when shift-clicking items)

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/game/rule/GameRule.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/rule/GameRule.java
@@ -20,7 +20,8 @@ public final class GameRule {
     public static final GameRule DISMOUNT_VEHICLE = new GameRule();
     public static final GameRule PLAYER_PROJECTILE_KNOCKBACK = new GameRule();
     public static final GameRule TRIDENTS_LOYAL_IN_VOID = new GameRule();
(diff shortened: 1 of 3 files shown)


Gegy (approved):
Looks good to me. Thank you! :)

Geek202:
Just looked at this again and thought that there should probably be a seperate rule for armour slots, as games may only want to lock the armour but be able to move stuff in their inventory

haykam821:
The `MODIFY_ARMOUR` rule and related names should be changed to `MODIFY_ARMOR`.

haykam821 (changes requested):
The `MODIFY_ARMOUR` rule should be renamed to `MODIFY_ARMOR` to match the Yarn and American English spelling.

Geek202:
Done!

Gegy:
It might make sense for MODIFY_INVENTORY to include MODIFY_ARMOR's behavior, given it might be confusing to disable inventory mutation but armor still works. I'm not sure if there's ever a use case for being able to modify armor but not the inventory- if there is, maybe that could work by inventory disallowed and armor explicitly allowed?

Gegy (approved):

PR #135, opened 2021-04-29
Add class to search for blocks around

Hugman76:
In a similar fashion as a bucket tool in painting programs, this class allow to detect all the blocks of a tag that are directly connected to a first block position. This allows for fast tree/ore veins destroying.

(PORTED from [BwActive](https://github.com/NucleoidMC/bed-wars/blob/1.16/src/main/java/xyz/nucleoid/bedwars/game/active/BwActive.java))

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
+++ b/src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
@@ -1,43 +1,44 @@
 package xyz.nucleoid.plasmid.util;
 
-import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.server.world.ServerWorld;
-import net.minecraft.tag.Tag;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Direction;
 
 import java.util.HashSet;
 import java.util.Set;
+import java.util.function.Predicate;
 
-public class BucketFind {
+public final class BucketFind {
     /**
-     * Finds any near blocks and puts them in a set.
+     * Finds any directly connected blocks and puts them in a set.
      *
-     * @param pos      The position of the first block
-     * @param blockTag The tag of the blocks that can be accepted in the set
+     * @param origin The position of the first block
+     * @param predicate The predicate for the blocks that can be accepted in the set
+     * @param limit The amount of maximum blocks to find
      */
-    public static Set<BlockPos> find(ServerWorld world, BlockPos pos, Tag<Block> blockTag) {
-        Set<BlockPos> positions = new HashSet<>();
-        positions.add(pos);
-        searchAround(world, pos, positions, blockTag);
-        return positions;
-    }
-
-    private static void searchAround(ServerWorld world, BlockPos pos, Set<BlockPos> positions, Tag<Block> blockTag) {
-        for(int x = -1; x <= 1; x++) {
-            for(int z = -1; z <= 1; z++) {
-                for(int y = -1; y <= 1; y++) {
-                    BlockPos local = pos.add(x, y, z);
-                    BlockState state = world.getBlockState(local);
-
-                    if(!positions.contains(local)) {
-                        if(state.isIn(blockTag)) {
-                            positions.add(local);
-                            searchAround(world, local, positions, blockTag);
-                        }
+    public static Set<BlockPos> find(ServerWorld world, BlockPos origin, Predicate<BlockState> predicate, int limit) {
+        Set<BlockPos> set = new HashSet<>();
+        Set<BlockPos> ends = new HashSet<>();
+        ends.add(origin);
+        while(limit > 0) {
+            if(ends.isEmpty()) {
+                return set;
+            }
+            BlockPos pos = ends.stream().findAny().get();
+            for(Direction direction : Direction.values()) {
+                BlockPos.Mutable local = pos.offset(direction).mutableCopy();
+                BlockState state = world.getBlockState(local);
+                if(predicate.test(state)) {
+                    if(!set.contains(local)) {
+                        ends.add(local);
                     }
                 }
             }
+            set.add(pos);
+            ends.remove(pos);
+            limit--;
         }
+        return set;
     }
 }
--- a/src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
+++ b/src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
@@ -1,5 +1,7 @@
 package xyz.nucleoid.plasmid.util;
 
+import it.unimi.dsi.fastutil.longs.LongArraySet;
+import it.unimi.dsi.fastutil.longs.LongSet;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.server.world.ServerWorld;
@@ -14,7 +16,7 @@ import java.util.Set;
 import java.util.function.Predicate;
 
 /**
- * Methods for finding connected blocks.
+ * Methods for finding connected blocks as a {@link LongSet}. Use {@link BlockPos#fromLong} to retrieve these positions.
  *
  * @see <a href="https://en.wikipedia.org/wiki/Pixel_connectivity#3-dimensional">3-dimensional pixel connectivity</a>
  */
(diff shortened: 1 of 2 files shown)


src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
```diff
@@ -0,0 +1,43 @@
...
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.tag.Tag;
+import net.minecraft.util.math.BlockPos;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class BucketFind {
+    /**
+     * Finds any near blocks and puts them in a set.
```
Restioson, line 14: This is insufficiently clear that it only finds connected blocks

Gegy:
Thank you for the PR! I think given we’re putting this in Plasmid we may want to look at extending the API a bit- like maybe supporting taking a Predicate? Also returning a LongSet might be good for allocation purposes.

Another thing which isn’t as important since it doesn’t change the API surface would be to use a deque instead of recursion such as to not blow up the stack. 😁 Maybe even a maximum search depth parameter?

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
```diff
@@ -0,0 +1,43 @@
...
+        positions.add(pos);
+        searchAround(world, pos, positions, blockTag);
+        return positions;
+    }
+
+    private static void searchAround(ServerWorld world, BlockPos pos, Set<BlockPos> positions, Tag<Block> blockTag) {
+        for(int x = -1; x <= 1; x++) {
+            for(int z = -1; z <= 1; z++) {
+                for(int y = -1; y <= 1; y++) {
+                    BlockPos local = pos.add(x, y, z);
```
haykam821, line 30: Use `BlockPos.Mutable` here.

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
```diff
@@ -0,0 +1,43 @@
...
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.tag.Tag;
+import net.minecraft.util.math.BlockPos;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class BucketFind {
```
haykam821, line 12: Since this is a utility class, it should not be extendable or constructable:

```suggestion
public final class BucketFind {
    private BucketFind() {}

```

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
```diff
@@ -0,0 +1,43 @@
...
+        Set<BlockPos> positions = new HashSet<>();
+        positions.add(pos);
+        searchAround(world, pos, positions, blockTag);
+        return positions;
+    }
+
+    private static void searchAround(ServerWorld world, BlockPos pos, Set<BlockPos> positions, Tag<Block> blockTag) {
+        for(int x = -1; x <= 1; x++) {
+            for(int z = -1; z <= 1; z++) {
+                for(int y = -1; y <= 1; y++) {
```
haykam821, line 29: I think this should only search the 6 blocks directly connected to `pos`. Iterating over `Direction.values()` instead should work better.

haykam821 (changes requested):

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
```diff
@@ -0,0 +1,44 @@
...
+        Set<BlockPos> set = new HashSet<>();
+        Set<BlockPos> ends = new HashSet<>();
+        ends.add(origin);
+        while(limit > 0) {
+            if(ends.isEmpty()) {
+                return set;
+            }
+            BlockPos pos = ends.stream().findAny().get();
+            for(Direction direction : Direction.values()) {
+                BlockPos.Mutable local = pos.offset(direction).mutableCopy();
```
haykam821, line 30: Extract the `BlockPos.Mutable` from the loop, and reuse a single instance.

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
```diff
@@ -0,0 +1,44 @@
...
+
+public final class BucketFind {
+    /**
+     * Finds any directly connected blocks and puts them in a set.
+     *
+     * @param origin The position of the first block
+     * @param predicate The predicate for the blocks that can be accepted in the set
+     * @param limit The amount of maximum blocks to find
+     */
+    public static Set<BlockPos> find(ServerWorld world, BlockPos origin, Predicate<BlockState> predicate, int limit) {
```
haykam821, line 20: An additional `findByTag` method which allows passing a `Tag<Block>` rather than a `Predicate<BlockState>` would be helpful.

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Set;
+import java.util.function.Predicate;
+
+public final class BucketFind {
+    /**
+     * Finds any directly connected blocks and puts them in a set.
+     *
+     * @param origin The position of the first block
+     * @param predicate The predicate for the blocks that can be accepted in the set
+     * @param limit The amount of maximum blocks to find
```
haykam821, line 18:
```suggestion
     * @param origin the position of the first block
     * @param predicate the predicate for the blocks that can be accepted in the set
     * @param limit the amount of maximum blocks to find
```

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
haykam821, line 20: `depth` would be more useful than `limit`, though both can be kept.

haykam821 (changes requested):

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
haykam821: Fixed in 3145459e790e7740817617dd1e6999ad831ae704

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
haykam821: Fixed in 3145459e790e7740817617dd1e6999ad831ae704

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
haykam821: This was only partially fixed in 3145459e790e7740817617dd1e6999ad831ae704; in addition to the `final` modifier on the class, the constructor should be made private as well.

src/main/java/xyz/nucleoid/plasmid/util/BucketFind.java
Hugman76: Is it better in e00ec75575a0bbea010b766f1e21dbdaf7f06de7?

haykam821: No, because you are now just assigning multiple `BlockPos.Mutable` instances to the same local variable.

Hugman76: Forgot about `BlockPos.Mutable#move` 😅 is it good in 6d9e3f3?

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,231 @@
...
+     * @param origin    the position of the first block
+     * @param depth     the amount of maximum blocks to find
+     * @param predicate the predicate for the blocks that can be accepted in the set
+     */
+    public static LongSet findEighteen(BlockPos origin, int depth, Predicate<BlockPos> predicate) {
+        LongSet set = new LongArraySet();
+        Deque<BlockPos> ends = new ArrayDeque<>();
+        ends.push(origin);
+        BlockPos.Mutable local = origin.mutableCopy();
+        while(depth > 0) {
```
Gegy, line 132: It seems like this isn't depth but rather just a count- I imagine this should be depth in the sense of how many steps have been taken from the source block?

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,231 @@
...
+            if(ends.isEmpty()) {
+                return set;
+            }
+            BlockPos pos = ends.pollLast();
+            for(Direction direction : Direction.values()) {
```
Gegy, line 75: It seems there's a bit of duplication with the logic that performs the search- potentially this type could be an enum for each neighbor handling method, and each implementation implements that neighbor logic specifically?

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
Hugman76: Yeah I guess count would make more sense here

Gegy: Count would be more accurate to the existing implementation but depth is more useful

Hugman76: Hmm.. what about `amount`? Also, why would depth be more useful?

Gegy: It’s hard to relate a count to what the method will scan- especially since it depends heavily on the implementation. If it’s breadth-first or depth-first, or what order it iterates neighbours, very different results will be selected.

Hugman76: Oh, alright, well I could implement both ideas with a simple boolean

Hugman76: Ugh, actually I think I'll stick with depth-first

Hugman76: Added both methods in e6cfa44 😄

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
Hugman76: Done in e6cfa44! 😄

Hugman76:
I switched the Predicate with a BiPredicate with the previous pos and next pos to allow for more possibilities

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -28,7 +28,7 @@ private BucketScanner() {
      * @param amount the amount of maximum blocks to find
      * @param predicate the predicate for the blocks that can be accepted in the set
```
haykam821, line 41: I would recommend documenting the purpose of the two `BlockPos` type parameters of the `BiPredicate`.

haykam821 (changes requested):

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
Gegy: Could alternatively use a custom functional interface here with named parameters

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,253 @@
...
+     * @return the set of positions found by the scanner that verify the predicate
+     */
+    public LongSet getPositions(Predicate<Integer> predicate) {
+        return new LongArraySet(depthMap.entrySet().stream().filter(entry -> predicate.test(entry.getValue())).mapToLong(Map.Entry::getKey).boxed().collect(Collectors.toSet()));
+    }
+
+    public enum Connectivity {
+        SIX,
+        EIGHTEEN,
+        TWENTY_SIX
```
haykam821, line 161: Document which blocks are effected by each connectivity here.

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,253 @@
...
+        return new LongArraySet(depthMap.keySet());
+    }
+
+    /**
+     * Gives a set of positions depending on the depth values of those positions.
+     *
+     * @param predicate determines if a position should enter the set depending of its depth value
+     * @return the set of positions found by the scanner that verify the predicate
+     */
+    public LongSet getPositions(Predicate<Integer> predicate) {
```
haykam821, line 154: I would use an `IntPredicate` or even a custom functional interface here:

```suggestion
    public LongSet getPositions(IntPredicate predicate) {
```

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,253 @@
...
+            long pos;
+            switch(builder.searchType) {
+                case BREADTH_FIRST:
+                default:
+                    pos = scanQueue.pollLast();
+                    break;
+                case DEPTH_FIRST:
+                    pos = scanQueue.pollFirst();
+                    break;
+            }
```
haykam821, line 98: This switch statement could be replaced by a field on the `SearchType` enum.

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,253 @@
...
+        return new Builder(predicate, Connectivity.SIX, SearchType.DEPTH_FIRST, 512, -1);
+    }
+
+    /**
+     * Creates a bucket scanner builder.
+     *
+     * @param block the block type that can be accepted in the set
+     * @return a new bucket scanner builder
+     */
+    public static Builder create(Block block, ServerWorld world) {
```
haykam821, line 59: I would recommend adding an additional method that has a `BlockState` parameter.

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        /**
+         * The maximum amount of blocks that can be scanned.
+         * <p>Set to -1 to disable this.
+         * <p>Is set to -1 by default.
+         *
+         * @return this builder for chaining
+         */
+        public Builder maxAmount(int max) {
```
haykam821, line 202: Name this parameter like the method and field: `maxAmount`.

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        /**
+         * The maximum length of branches that can be scanned.
+         * <p>Cannot be set to a value under 0.
+         * <p>Is set to 512 by default.
+         *
+         * @return this builder for chaining
+         */
+        public Builder maxDepth(int max) {
```
haykam821, line 214: Name this parameter like the method and field: `maxDepth`.

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        /**
+         * The type of connectivity to use to search blocks around a branch end.
+         * Is set to {@link Connectivity#SIX} by default.
+         *
+         * @return this builder for chaining
+         * @see <a href="https://en.wikipedia.org/wiki/Pixel_connectivity#3-dimensional">3-dimensional pixel connectivity</a>
+         */
+        public Builder connectivity(Connectivity c) {
```
haykam821, line 226: Name this parameter like the method and field: `connectivity`.

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        /**
+         * Is set to {@link SearchType#DEPTH_FIRST} by default.
+         *
+         * @return this builder for chaining
+         * @see <a href="https://en.wikipedia.org/wiki/Breadth-first_search">Breadth-first search</a>
+         * @see <a href="https://en.wikipedia.org/wiki/Depth-first_search">Depth-first search</a>
+         */
+        public Builder searchType(SearchType s) {
```
haykam821, line 238: Name this parameter like the method and field: `searchType`.

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
```diff
@@ -0,0 +1,253 @@
...
+
+    public enum SearchType {
+        BREADTH_FIRST,
+        DEPTH_FIRST
+    }
+
+    @FunctionalInterface
+    public interface ScanPredicate {
+        /**
+         * Method executed at each branch end to fuel the scanning process and spread around.
```
haykam821, line 172: This summary is confusing, since a predicate isn't a furnace. Could verbs better than 'fuel' and 'spread' be used here?

haykam821 (changes requested):

src/main/java/xyz/nucleoid/plasmid/util/BucketScanner.java
Hugman76: I think another custom functional interface might be too much, just for an integer predicate.

PR #136, opened 2021-04-29
Add WoodBlocks

Hugman76:
(PORTED from [Bedwars](https://github.com/NucleoidMC/bed-wars/blob/1.16/src/main/java/xyz/nucleoid/bedwars/util/WoodBlocks.java))


haykam821:
Perhaps this should be extracted into a library along with wool colors?

Hugman76:
I don't know, this seems like too little content to make a new library?

src/main/java/xyz/nucleoid/plasmid/util/WoodBlocks.java
```diff
@@ -0,0 +1,31 @@
+package xyz.nucleoid.plasmid.util;
+
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.block.Blocks;
+
+public class WoodBlocks {
+    public static BlockState saplingOf(BlockState state) {
```
haykam821, line 8: There should be an additional method for block-to-block conversions. This method should instead copy the state's properties.

src/main/java/xyz/nucleoid/plasmid/util/WoodBlocks.java
```diff
@@ -0,0 +1,31 @@
...
+        Block block = state.getBlock();
+        if (block == Blocks.OAK_LEAVES) return Blocks.OAK_SAPLING.getDefaultState();
+        if (block == Blocks.BIRCH_LEAVES) return Blocks.BIRCH_SAPLING.getDefaultState();
+        if (block == Blocks.SPRUCE_LEAVES) return Blocks.SPRUCE_SAPLING.getDefaultState();
+        if (block == Blocks.JUNGLE_LEAVES) return Blocks.JUNGLE_SAPLING.getDefaultState();
+        if (block == Blocks.ACACIA_LEAVES) return Blocks.ACACIA_SAPLING.getDefaultState();
+        if (block == Blocks.DARK_OAK_LEAVES) return Blocks.DARK_OAK_SAPLING.getDefaultState();
+
+        return Blocks.AIR.getDefaultState();
```
haykam821, line 17: Would `null` be a better option?

haykam821 (changes requested):

src/main/java/xyz/nucleoid/plasmid/util/WoodBlocks.java
Restioson, line 17: If this is going to be upstream code then IMO it should be held to greater standards. For instance, this could rather raise some kind of exception, or the function could return Optional<BlockState>

src/main/java/xyz/nucleoid/plasmid/util/WoodBlocks.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class WoodBlocks {
+    public static BlockState saplingOf(BlockState state) {
+        Block block = state.getBlock();
+        if (block == Blocks.OAK_LEAVES) return Blocks.OAK_SAPLING.getDefaultState();
+        if (block == Blocks.BIRCH_LEAVES) return Blocks.BIRCH_SAPLING.getDefaultState();
+        if (block == Blocks.SPRUCE_LEAVES) return Blocks.SPRUCE_SAPLING.getDefaultState();
+        if (block == Blocks.JUNGLE_LEAVES) return Blocks.JUNGLE_SAPLING.getDefaultState();
+        if (block == Blocks.ACACIA_LEAVES) return Blocks.ACACIA_SAPLING.getDefaultState();
+        if (block == Blocks.DARK_OAK_LEAVES) return Blocks.DARK_OAK_SAPLING.getDefaultState();
```
Restioson, line 15: Replacing this with a switch-case statement is probably best

Hugman76:
I've tried making an enum instead. I thought it would be better like that

src/main/java/xyz/nucleoid/plasmid/util/WoodType.java
```diff
@@ -0,0 +1,104 @@
...
+        this.door = door;
+        this.sign = sign;
+        this.button = button;
+        this.pressurePlate = pressurePlate;
+        this.boat = boat;
+    }
+
+    public WoodType getType(Block block) {
+        for (WoodType type : WoodType.values()) {
+            if (block == type.sapling || block == type.leaves || block == type.log || block == type.planks || block == type.slab || block == type.stairs || block == type.fence || block == type.fenceGate || block == type.door || block == type.sign || block == type.button || block == type.pressurePlate) {
```
Gegy, line 37: This might be nice as a `.contains(Block)` method

Hugman76: Done 😄

Gegy (approved):

PR #138, opened 2021-05-07
Replace fake items implementation with polymer

Patbox:
This pr replaces plasmid fake items implementation with one from [polymer](https://github.com/Patbox/polymer) and deprecates old one.

I've tested it with unmodified Fortress and Bedwars and I didn't find any issues.

Additionally plasmid will now bundle [sgui](https://github.com/Patbox/sgui) for simpler gui creation.


arthurbambou (approved):

build.gradle
```diff
@@ -51,6 +51,12 @@ dependencies {
   include 'xyz.nucleoid:fantasy:0.2.6'
 
   modImplementation 'xyz.nucleoid:leukocyte:0.1.3'
+
+  modImplementation 'com.github.Patbox:polymer:0.0.1'
+  include 'com.github.Patbox:polymer:0.0.1'
+
+  modImplementation 'com.github.Patbox:sgui:0.3.0'
+  include 'com.github.Patbox:sgui:0.3.0'
```
haykam821, line 59: I don't think it's necessary to bundle something that isn't used.

src/main/java/xyz/nucleoid/plasmid/fake/FakeBlock.java
```diff
@@ -28,6 +30,7 @@ default FluidState asProxy(FluidState state) {
         return this.asProxy(state.getBlockState()).getFluidState();
     }
 
+
```
haykam821, line 33: An unnecessary empty line was added here.

src/main/java/xyz/nucleoid/plasmid/fake/FakeEntityType.java
```diff
@@ -4,6 +4,7 @@
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityType;
 
+@Deprecated
```
haykam821, line 7: Is there an actual replacement for fake entity types?

src/main/java/xyz/nucleoid/plasmid/item/AddRegionItem.java
```diff
@@ -40,7 +42,14 @@ public AddRegionItem(Item.Settings settings) {
...
+    public Item getVirtualItem() {
         return Items.STICK;
     }
+
+    @Override
+    public ItemStack getVirtualItemStack(ItemStack itemStack, ServerPlayerEntity player) {
+        ItemStack virtual = ItemHelper.createBasicVirtualItemStack(itemStack, player);
+        virtual.addEnchantment(Enchantments.POWER, 1);
+        return virtual;
+    }
```
haykam821, line 54: I don't think the proxy stack should be changed.

src/main/java/xyz/nucleoid/plasmid/item/IncludeEntityItem.java
```diff
@@ -64,7 +67,14 @@ public ActionResult useOnEntity(ItemStack stack, PlayerEntity user, LivingEntity
...
-        return Items.DEBUG_STICK;
+    public Item getVirtualItem() {
+        return Items.LEAD;
+    }
+
+    @Override
+    public ItemStack getVirtualItemStack(ItemStack itemStack, ServerPlayerEntity player) {
+        ItemStack virtual = ItemHelper.createBasicVirtualItemStack(itemStack, player);
+        virtual.addEnchantment(Enchantments.POWER, 1);
+        return virtual;
```
haykam821, line 78: I don't think the proxy stack should be changed here either.

haykam821 (changes requested):

src/main/java/xyz/nucleoid/plasmid/fake/FakeBlock.java
```diff
@@ -1,13 +1,15 @@
...
+import eu.pb4.polymer.block.VirtualBlock;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.fluid.FluidState;
 
 /**
  * Represents a block that should be remapped to some vanilla "proxy" counterpart when being sent to clients
  */
-public interface FakeBlock {
+@Deprecated
```
Gegy, line 11: Can we add documentation to all these deprecated classes to rather point towards the correct replacement? 🙂

Patbox: Something like 
```
/**
 * @deprecated This interface is deprecated in favour of {@link eu.pb4.polymer.block.VirtualBlock}
 *
 * Represents a block that should be remapped to some vanilla "proxy" counterpart when being sent to clients
 */
``` should work?

PR #141, opened 2021-05-15
Allow writing an instance of the Text class to the content of a sidebar

haykam821:


Gegy:
This is now supported since using Patbox's sidebar API :)

PR #150, opened 2021-05-25
Block traversal API

Gegy:
Supersedes #135


Hugman76:
Great! Thanks a lot for the help <3

PR #153, opened 2021-06-25
Preliminary 0.5 refactors

Gegy:
These are the preliminary refactors to 0.5 which quite significantly change the way mods interact with Plasmid. This includes:
 - Migrate to use of Stimuli for events
 - Simplify game channels and replace with "portals"
 - Update to Fantasy 0.4: Plasmid is now responsible for handling player isolation
 - Allow for GameSpaces to have multiple worlds attached

This is very much still WIP and will absolutely change with feedback & testing of 0.5- so it's not set in stone at all, but for now creating this PR for getting early feedback. 🙂 

In terms of the following roadmap for going forward with 0.5:
 - Get basic approval on these initial refactors from contributors (not finalised at all!)
 - Update Yarn
 - Update to 1.17
 - Migrate to various external libraries where relevant (sgui, holograms) + many other refactor areas
 - Work on documenting all APIs
 - Start publishing builds of 0.5 for mods to start testing and updating to
 - Get feedback from minigame developers & iterate until we are all happy with 0.5
 - Release when we feel it's ready!


src/main/resources/data/plasmid/lang/fr_fr.json
```diff
@@ -2,20 +2,19 @@
   "game.plasmid.random": "Jeu aléatoire",
   "item.plasmid.add_region": "Ajouter une région",
   "item.plasmid.include_entity": "Inclure une entité",
-  "text.plasmid.channel.channel_not_found": "Ce canal n'a pas été trouvé !",
-  "text.plasmid.channel.join_link_hover": "Rejoindre %s",
+  "text.plasmid.portal.portal_not_found": "Ce canal n'a pas été trouvé !",
```
LambdAurora, line 5:
```suggestion
  "text.plasmid.portal.portal_not_found": "Ce portail n'a pas été trouvé !",
```

src/main/java/xyz/nucleoid/plasmid/entity/FloatingText.java
```diff
@@ -43,7 +43,7 @@ public void setText(Text line) {
         this.setText(new Text[] { line });
     }
 
-    // TODO: 0.5- api needs to better inform caller whether spawning failed
+    // TODO: 0.5- migrate to patbox's floating text api
```
LambdAurora, line 46: The question is, do we want to just rebuild the implementation of this on top of that Holograms API or do we just deprecate and encourage the use of the Holograms API instead?

src/main/java/xyz/nucleoid/plasmid/game/player/PlayerOffer.java
```diff
@@ -0,0 +1,30 @@
+package xyz.nucleoid.plasmid.game.player;
+
+import net.minecraft.server.network.ServerPlayerEntity;
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.text.Text;
+import net.minecraft.util.math.Vec3d;
+
+public final class PlayerOffer {
```
LambdAurora, line 8: Would like some documentation, currently I have no idea what this does.

src/main/java/xyz/nucleoid/plasmid/game/player/isolation/IsolatingPlayerTeleporter.java
```diff
@@ -0,0 +1,103 @@
...
+import net.minecraft.server.PlayerManager;
+import net.minecraft.server.network.ServerPlayNetworkHandler;
+import net.minecraft.server.network.ServerPlayerEntity;
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.world.WorldProperties;
+import net.minecraft.world.biome.source.BiomeAccess;
+
+import java.util.function.Function;
+
+public final class IsolatingPlayerTeleporter {
```
LambdAurora, line 15: Documentation?

src/main/java/xyz/nucleoid/plasmid/game/player/isolation/PlayerResetter.java
```diff
@@ -0,0 +1,43 @@
...
+import net.minecraft.entity.attribute.AttributeContainer;
+import net.minecraft.entity.attribute.EntityAttribute;
+import net.minecraft.entity.attribute.EntityAttributeInstance;
+import net.minecraft.entity.attribute.EntityAttributeModifier;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.server.network.ServerPlayerEntity;
+import net.minecraft.util.registry.Registry;
+
+public final class PlayerResetter {
+    private final CompoundTag resetTag;
```
LambdAurora, line 12:
```suggestion
    private final CompoundTag resetNbt;
```

LambdAurora (commented):
I highly recommend to update to the latest yarn mappings to have the backported mappings from 1.17 (where it was applicable), which would ease a bit the 1.17 port, and it contains the NBT refactors.

src/main/java/xyz/nucleoid/plasmid/entity/FloatingText.java
Gegy: I think we can rather just include the holograms api- since this is a breaking version anyway, we may as well remove deprecated stuff directly.

Gegy:
@LambdAurora Regarding updating Yarn: absolutely need to do this before porting- want to leave it for now until we have approval on this PR though so the diff is *slightly* easier to review. 😄 

Regarding documentation: absolutely! There's a lot of documentation work I need to get done- for some things I want to wait a bit before documenting given the API might still change significantly, but for the stuff that will *probably* stay the same, I'm going to go through and add some docs. 🙂

LambdAurora:
Also, I saw some `isClient` calls, I would use the method instead of the variable imo, especially since Mojang has a tendency to hide variables and use methods instead.

PR #154, opened 2021-06-27
Statistics API

### LINKED ISSUE #105: Statistics API

Tom-The-Geek:
Closes #105

Notes:
- This API doesn't actually store any values after the game ends, it would be the responsibility of another mod (eg. `nucleoid-extras`) to implement the communication with a storage backend and capture the `GameStatisticBundle` using `GameEvents.CLOSING`

Questions:
- Should plasmid provide some default `StatisticKeys` (eg. kills, wins, etc) for minigames to use, and if so, what should be included?
  - Yes
- Should plasmid provide a method of querying statistics after a game has ended, or should that be left up to the storage implementation (eg. `nucleoid-extras`) to provide an API?
  - Later:tm:


src/main/java/xyz/nucleoid/plasmid/game/stats/GameStatisticBundle.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public StatisticBundle getPlayer(UUID uuid) {
+        if (!this.players.containsKey(uuid)) {
+            this.players.put(uuid, new StatisticBundle());
+        }
+        return this.players.getOrDefault(uuid, new StatisticBundle());
+    }
+
+    // Gotta love generics
+    public void increment(ServerPlayerEntity player, StatisticKey<Double> key, double amount) {
```
Gegy, line 34: It's probably fine to keep these functions just on the `StatisticsBundle`, as `getPlayer(..).increment(..)` seems more clear in any case.

Also means a lot less duplication for each type of player parameter. 😄

src/main/java/xyz/nucleoid/plasmid/game/stats/GameStatisticBundle.java
```diff
@@ -0,0 +1,81 @@
...
+import xyz.nucleoid.plasmid.util.PlayerRef;
+
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * Wrapper containing a map players' {@link StatisticBundle}s
+ */
+public class GameStatisticBundle {
+    private final Object2ObjectMap<UUID, StatisticBundle> players = new Object2ObjectOpenHashMap<>();
```
Gegy, line 16: Could be interesting to also provide a method for storing global statistics- things such as total games played / average win rates for certain teams etc. If games can retrieve statistics for themselves, they could potentially do interesting things with that data- or it could be used for balancing.

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
```diff
@@ -0,0 +1,84 @@
...
+     *
+     * @param id The {@link Identifier} to check
+     */
+    private static void validateKeyId(Identifier id) {
+        if (id.getNamespace().contains(".") || id.getPath().contains(".")) {
+            throw new IllegalArgumentException("StatisticKey ids cannot contain '.'");
+        }
+    }
+
+    public enum StatisticType implements StringIdentifiable {
```
Gegy, line 69: Maybe this could be `StoreType`? Or something along those lines to indicate it controls how the statistic is stored/handled.

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public JsonObject encodeValue(T value) {
+        JsonObject obj = new JsonObject();
+        obj.addProperty("type", this.encodeType());
+        obj.addProperty("value", value);
+        return obj;
+    }
+
+    protected abstract String encodeType();
```
Gegy, line 44: I think given this is the only abstract function here, this class probably doesn't need to be abstract- maybe it should instead hold two types like a data type (float/int/etc) and a store type (rolling avg/total)?

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
```diff
@@ -0,0 +1,84 @@
...
+import net.minecraft.util.Identifier;
+import net.minecraft.util.StringIdentifiable;
+import xyz.nucleoid.plasmid.game.stats.key.DoubleStatisticKey;
+import xyz.nucleoid.plasmid.game.stats.key.FloatStatisticKey;
+import xyz.nucleoid.plasmid.game.stats.key.IntStatisticKey;
+
+/**
+ * Type-safe keys for identifying statistics.
+ * <p>
+ * Ideally should have a corresponding translation key in the form <code>statistic.[bundle namespace].[key namespace].[key path]</code>
```
Gegy, line 13: Should statistic translation keys involve the bundle namespace? I imagine the same statistic key would always mean the same thing irrespective of what game it was involved in?

Gegy:
Regarding your two questions:
1. I think it definitely should provide some defaults- potentially like: `kills`, `deaths`, `games_played`, `games_won`, `games_lost`, `points`, `placement`, `time_survived`, `damage_taken`, `damage_dealt`, `blocks_broken` as an initial few? We can always add more later.
2. It could definitely be nice for there to be a Plasmid API to query statistics through, but potentially we leave that for a later stage of implementation? Definitely needs some thinking about in terms of how we want to handle implementing it.

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
Tom-The-Geek: On the line below it says that the translation key can also be `statistic.[key namespace].[key path]`, which could be used as a fallback if there is not a `statistic.[bundle namespace].[key namespace].[key path]` translation.
This would allow a game like bedwars to potentially store two bundles, such as `bedwars` and `bedwars-teams`, both containing a value for the statistic `plasmid:wins`, but can override the display name for the `bedwars-teams` bundle value to be `Wins (teams)` without requiring another statistic key

src/main/java/xyz/nucleoid/plasmid/game/stats/GameStatisticBundle.java
Tom-The-Geek: This would also require changes to the backend for storage too, but I can add an additional field to `GameStatisticBundle` for global statistics so that an API can be exposed in plasmid

Tom-The-Geek:
I'll take a look at implementing the suggestions after I've eaten, and yes I agree that querying statistics through plasmid should probably be left for now, so that it can be thought about.

src/main/java/xyz/nucleoid/plasmid/game/stats/GameStatisticBundle.java
```diff
@@ -0,0 +1,57 @@
...
+    private final StatisticBundle global = new StatisticBundle();
```
Gegy, line 17: `GameStatisticsBundle` vs `StatisticBundle` is a bit of a confusing distinction- what about renaming `StatisticBundle` -> `StatisticMap` or similar?

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticBundle.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.google.gson.JsonObject;
+import it.unimi.dsi.fastutil.objects.Object2ObjectMap;
+import it.unimi.dsi.fastutil.objects.Object2ObjectOpenHashMap;
+
+/**
+ * Stores a mapping of {@link StatisticKey} to their corresponding values
+ */
+public class StatisticBundle {
+    private final Object2ObjectMap<StatisticKey<?>, Number> values = new Object2ObjectOpenHashMap<>();
```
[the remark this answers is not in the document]
Tom-The-Geek: That could be useful for statistics such as ranking, where the game doesn't want to have to loop to increment to the required value.

src/main/java/xyz/nucleoid/plasmid/game/stats/GameStatisticBundle.java
```diff
@@ -0,0 +1,57 @@
...
+    public boolean isEmpty() {
+        return this.global.isEmpty()
+                && (this.players.isEmpty() || this.players.values().stream().allMatch(StatisticBundle::isEmpty));
+    }
+
+    public JsonObject encodeBundle() {
+        JsonObject obj = new JsonObject();
+        JsonObject players = new JsonObject();
+        for (Map.Entry<UUID, StatisticBundle> entry : this.players.entrySet()) {
+            players.add(entry.getKey().toString(), entry.getValue().encodeBundle());
```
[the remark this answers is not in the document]
Tom-The-Geek: Missed that, thanks!

src/main/java/xyz/nucleoid/plasmid/game/stats/GameStatisticBundle.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public StatisticBundle getPlayer(ServerPlayerEntity player) {
+        return this.getPlayer(player.getUuid());
+    }
+
+    public StatisticBundle getPlayer(UUID uuid) {
+        if (!this.players.containsKey(uuid)) {
+            this.players.put(uuid, new StatisticBundle());
+        }
+        return this.players.getOrDefault(uuid, new StatisticBundle());
```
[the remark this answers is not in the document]
Tom-The-Geek: I have a feeling I forgot what that method was called when I wrote this :P

src/main/java/xyz/nucleoid/plasmid/game/stats/GameStatisticBundle.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.UUID;
+
+/**
+ * Wrapper containing a map players' {@link StatisticBundle}s
+ */
+public class GameStatisticBundle {
+    private final Object2ObjectMap<UUID, StatisticBundle> players = new Object2ObjectOpenHashMap<>();
+    private final StatisticBundle global = new StatisticBundle();
+
+    public StatisticBundle getPlayer(PlayerRef player) {
```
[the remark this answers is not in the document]
Tom-The-Geek: This definitely reads better after having tried the names, eg.
```java
statistics.forPlayer(player).increment(TEST_KEY, 2.5);
```

Tom-The-Geek (commented):
I'll push changes for these in a second :+1:

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticMap.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public void increment(StatisticKey<Float> key, float amount) {
+        this.values.put(key, (Float) this.values.getOrDefault(key, 0.0f) + amount);
+    }
+
+    public void increment(StatisticKey<Integer> key, int amount) {
+        this.values.put(key, (Integer) this.values.getOrDefault(key, 0) + amount);
+    }
+
+    public double getDouble(StatisticKey<Double> key) {
```
Gegy, line 25: Ahh erased generics 😆 
Might be best to name these all consistently with type suffix since we can't avoid it here

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
```diff
@@ -0,0 +1,104 @@
...
+        @Override
+        public String asString() {
+            return this.name;
+        }
+    }
+
+    public enum StorageType implements StringIdentifiable {
+        TOTAL("total"),
+        ROLLING_AVERAGE("rolling_average"),
+        ;
```
Gegy, line 92: Minimum/maximum storage tracking could be really useful to include also for stuff like fastest times

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKeys.java
```diff
@@ -0,0 +1,35 @@
...
+    public static final StatisticKey<Integer> DEATHS = StatisticKey.intKey(id("deaths"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Double> DAMAGE_TAKEN = StatisticKey.doubleKey(id("damage_taken"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Double> DAMAGE_DEALT = StatisticKey.doubleKey(id("damage_dealt"), StatisticKey.StorageType.TOTAL);
+
+    // Misc;
+    public static final StatisticKey<Integer> BLOCKS_BROKEN = StatisticKey.intKey(id("blocks_broken"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> POINTS = StatisticKey.intKey(id("points"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> TIME_SURVIVED = StatisticKey.intKey(id("time_survived"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> RANKING = StatisticKey.intKey(id("ranking"), StatisticKey.StorageType.TOTAL);
+
```
Gegy, line 30: I think we might want two time statistics where one is tracked by minimum and the other by maximum

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKeys.java
```diff
@@ -0,0 +1,35 @@
...
+    public static final StatisticKey<Integer> KILLS = StatisticKey.intKey(id("total_kills"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> DEATHS = StatisticKey.intKey(id("deaths"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Double> DAMAGE_TAKEN = StatisticKey.doubleKey(id("damage_taken"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Double> DAMAGE_DEALT = StatisticKey.doubleKey(id("damage_dealt"), StatisticKey.StorageType.TOTAL);
+
+    // Misc;
+    public static final StatisticKey<Integer> BLOCKS_BROKEN = StatisticKey.intKey(id("blocks_broken"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> POINTS = StatisticKey.intKey(id("points"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> TIME_SURVIVED = StatisticKey.intKey(id("time_survived"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> RANKING = StatisticKey.intKey(id("ranking"), StatisticKey.StorageType.TOTAL);
```
Gegy, line 29: Not sure why I suggested ranking if it only really has meaning within one game 😄  This is probably fine to remove

Gegy (commented):
Apologies for taking so long to review all the time! This is looking really good and just have some last very minor thoughts

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
Tom-The-Geek: I started work on this at the backend side, just didn't get round to pushing anything or implementing it plasmid side .

Tom-The-Geek:
Pushed those changes! I'm now going to take a look at the backend stuff again :smile:

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKeys.java
```diff
@@ -0,0 +1,35 @@
...
+    public static final StatisticKey<Integer> KILLS = StatisticKey.intKey(id("total_kills"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> DEATHS = StatisticKey.intKey(id("deaths"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Double> DAMAGE_TAKEN = StatisticKey.doubleKey(id("damage_taken"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Double> DAMAGE_DEALT = StatisticKey.doubleKey(id("damage_dealt"), StatisticKey.StorageType.TOTAL);
+
+    // Misc;
+    public static final StatisticKey<Integer> BLOCKS_BROKEN = StatisticKey.intKey(id("blocks_broken"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> POINTS = StatisticKey.intKey(id("points"), StatisticKey.StorageType.TOTAL);
+    public static final StatisticKey<Integer> MIN_TIME_SURVIVED = StatisticKey.intKey(id("time_survived"), StatisticKey.StorageType.MIN);
+    public static final StatisticKey<Integer> MAX_TIME_SURVIVED = StatisticKey.intKey(id("time_survived"), StatisticKey.StorageType.MAX);
```
Gegy, line 29: Hmm, maybe these should be named something like `longest_time`/`shortest_time`

Tom-The-Geek: Done!

arthurbambou (approved):

PR #155, opened 2021-07-01
Add the fluid flow rule

haykam821:
This pull request adds a fluid flow rule which is based on NucleoidMC/leukocyte#6.


Restioson:
Is there no way for plasmid to use leukocyte directly here?

haykam821:
There isn't a release of Leukocyte that can be used in 0.4 that has the fluid flow rule. In 0.5, this can be resolved.

PR #157, opened 2021-07-11
Implement TeamManager

Patbox:
This is more or less replacement for vanilla teams, it should work simpler and limit issues caused by games not removing teams from players

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/game/common/team/GameTeam.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/common/team/GameTeam.java
@@ -138,7 +138,7 @@ public final class GameTeam {
                     Codec.STRING.fieldOf("dye").forGetter((t) -> t.dyeColor.toString()),
                     MoreCodecs.DYE_COLOR.fieldOf("block_dye").forGetter((t) -> t.blockDyeColor),
                     Codec.STRING.fieldOf("firework").forGetter((t) -> t.fireworkColor.toString()),
-                    MoreCodecs.FORMATTING.fieldOf("formatting").forGetter((t) -> t.formatting)
+                    MoreCodecs.FORMATTING.optionalFieldOf("formatting", Formatting.RESET).forGetter((t) -> t.formatting)
             ).apply(instance, TeamColorData::new);
         });
 
--- a/src/main/java/xyz/nucleoid/plasmid/game/common/team/GameTeam.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/common/team/GameTeam.java
@@ -19,7 +19,7 @@ public final class GameTeam {
     public static final Codec<GameTeam> CODEC = RecordCodecBuilder.create(instance -> {
         return instance.group(
                 Codec.STRING.fieldOf("key").forGetter(GameTeam::key),
-                Codec.either(MoreCodecs.TEXT, Codec.STRING).fieldOf("display").forGetter((team) -> Either.left(team.display)),
+                MoreCodecs.TEXT.fieldOf("display").forGetter(GameTeam::display),
                 Codec.either(MoreCodecs.DYE_COLOR, TeamColorData.CODEC).fieldOf("color").forGetter((team) -> Either.right(new TeamColorData(team.color, team.dyeColor, team.blockDyeColor, team.fireworkColor, team.formatting)))
 
         ).apply(instance, GameTeam::new);
@@ -33,7 +33,15 @@ public final class GameTeam {
     private final Formatting formatting;
     private final DyeColor blockDyeColor;
 
-    public GameTeam(String key, Either<MutableText, String> display, Either<DyeColor, TeamColorData> color) {
+    public GameTeam(String key, MutableText display, DyeColor color) {
+        this(key, display, Either.left(color));
+    }
+
+    public GameTeam(String key, MutableText display, TeamColorData color) {
+        this(key, display, Either.right(color));
+    }
+
+    public GameTeam(String key, MutableText display, Either<DyeColor, TeamColorData> color) {
         this.key = key;
 
         if (color.left().isPresent()) {
@@ -52,7 +60,7 @@ public final class GameTeam {
             this.fireworkColor = colorData.fireworkColor;
         }
 
-        this.display = display.left().isPresent() ? display.left().get() : new TranslatableText(display.right().get()).setStyle(Style.EMPTY.withColor(this.color()));
+        this.display = display.styled((style) -> style.getColor() == null ? style.withColor(this.color) : style);
     }
 
     private static Formatting formatByDye(DyeColor dye) {
--- a/src/main/java/xyz/nucleoid/plasmid/game/common/team/GameTeam.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/common/team/GameTeam.java
@@ -1,9 +1,11 @@
 package xyz.nucleoid.plasmid.game.common.team;
 
+import com.mojang.datafixers.util.Either;
 import com.mojang.serialization.Codec;
 import com.mojang.serialization.codecs.RecordCodecBuilder;
 import net.minecraft.item.FireworkItem;
 import net.minecraft.item.ItemStack;
+import net.minecraft.text.*;
 import net.minecraft.util.DyeColor;
 import net.minecraft.util.Formatting;
 import net.minecraft.util.math.MathHelper;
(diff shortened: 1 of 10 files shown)


src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    public void setSuffix(GameTeam team, Text value) {
+        this.getVanilla(team).setSuffix(value);
+        this.sendUpdates(team);
+    }
+
+    public GameTeam getSmallestTeam() {
+        GameTeam smallest = null;
+        int count = 9999;
```
haykam821, line 279: I'm not sure if this is the best solution here, but either way it should use `Integer.MAX_VALUE` just to be safe.

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
```diff
@@ -0,0 +1,330 @@
...
+    public void disableFormatting() {
+        this.applyFormattingByDefault = false;
+    }
+
+    public void enableFormatting() {
+        this.applyFormattingByDefault = true;
+    }
+
+    /**
+     * Pre-registers team, allowing methods like {@link TeamManager#getSmallestTeam()} to work correctly
```
haykam821, line 301: It might be better to enforce team registration.

src/main/java/xyz/nucleoid/plasmid/game/common/team/GameTeam.java
```diff
@@ -13,40 +15,44 @@
...
  * A simple representation of a team type, containing a name and color.
  */
-public final record GameTeam(String key, String display, DyeColor dye, Formatting formatting) {
+public final class GameTeam {
     public static final Codec<GameTeam> CODEC = RecordCodecBuilder.create(instance -> {
         return instance.group(
                 Codec.STRING.fieldOf("key").forGetter(GameTeam::key),
-                Codec.STRING.fieldOf("display").forGetter(GameTeam::display),
-                MoreCodecs.DYE_COLOR.fieldOf("color").forGetter(GameTeam::dye)
+                Codec.either(MoreCodecs.TEXT, Codec.STRING).fieldOf("display").forGetter((team) -> Either.left(team.display)),
```
Gegy, line 22: I believe the text codec should already support parsing a string literal

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
```diff
@@ -0,0 +1,330 @@
...
+public final class TeamManager {
+    private final Map<GameTeam, Set<PlayerRef>> teamToPlayerRefSet = new Object2ObjectOpenHashMap<>();
+    private final Map<GameTeam, MutablePlayerSet> teamToPlayerSet = new Object2ObjectOpenHashMap<>();
+    private final Map<PlayerRef, GameTeam> playerRefToTeam = new Object2ObjectOpenHashMap<>();
+    private final Map<GameTeam, Team> teamToVanillaTeam = new Object2ObjectOpenHashMap<>();
+    private final Object2BooleanMap<GameTeam> teamFriendlyFire = new Object2BooleanArrayMap<>();
+
+    private final Scoreboard scoreboard = new Scoreboard();
+    private final MinecraftServer server;
+    private GameActivity gameActivity;
```
Gegy, line 40: The mutable state for the activity isn't ideal - is it needed? Could we avoid it by having `applyTo` return an initialized instance of a `TeamManager`?

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
```diff
@@ -0,0 +1,330 @@
...
+import xyz.nucleoid.plasmid.util.PlayerRef;
+import xyz.nucleoid.stimuli.event.player.PlayerDamageEvent;
+
+import java.util.*;
+
+/**
+ * Simple, GameActivity specific team manager class
+ */
+public final class TeamManager {
+    private final Map<GameTeam, Set<PlayerRef>> teamToPlayerRefSet = new Object2ObjectOpenHashMap<>();
```
Gegy, line 32: I'm a bit confused by the difference between `teamToPlayerRefSet` and `teamToPlayerSet` here- do they not serve the same purpose? They don't seem to be used consistently (e.g. in the player add listener)

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
```diff
@@ -0,0 +1,330 @@
...
+    public Text getSuffix(GameTeam team) {
+        return this.getVanilla(team).getSuffix();
+    }
+
+    public void setSuffix(GameTeam team, Text value) {
+        this.getVanilla(team).setSuffix(value);
+        this.sendUpdates(team);
+    }
+
+    public GameTeam getSmallestTeam() {
```
Gegy, line 277:
```suggestion
    @Nullable
    public GameTeam getSmallestTeam() {
```

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
Patbox: I could switch it technically for GameSpace, hovewer I prefer to keep it mutable so TeamManager can be passed to another GameActivity

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
```diff
@@ -0,0 +1,330 @@
...
+    public AbstractTeam.VisibilityRule getNameTagVisibilityRule(GameTeam team) {
+        return this.getVanilla(team).getNameTagVisibilityRule();
+    }
+
+    public void setNameTagVisibilityRule(GameTeam team, AbstractTeam.VisibilityRule rule) {
+        this.getVanilla(team).setNameTagVisibilityRule(rule);
+        this.sendUpdates(team);
+    }
+
+    public Text applyPlayerFormatting(PlayerRef player, Text text) {
```
Gegy, line 241: It isn't entirely clear what formatting this is or what it's being applied to- is this the player name?

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
```diff
@@ -0,0 +1,330 @@
...
+
+        activity.listen(GamePlayerEvents.ADD, (player) -> {
+            for (Team vanillaTeam : this.teamToVanillaTeam.values()) {
+                player.networkHandler.sendPacket(TeamS2CPacket.updateTeam(vanillaTeam, true));
+            }
+
+            PlayerRef ref = PlayerRef.of(player);
+            GameTeam team = this.playerRefToTeam.get(ref);
+            if (team != null) {
+                this.teamToPlayerSet.computeIfAbsent(team, (t) -> new MutablePlayerSet(this.server)).add(ref);
```
Gegy, line 63: Some duplication here along with the player team setters

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
Patbox: `teamToPlayerRefSet` contains set of all players that are added (including offline ones and ones that left the minigame).
`teamToPlayerSet` contains only players that are in the game

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
Gegy: Hmm, I believe it would not really be possible to pass between activities anyway given the player remove events and close events will be called. This is not really something that can be worked around at the moment, at least until we do the deeper refactors.

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
Gegy: `PlayerSet` is interesting since it can contain both online and offline players, but the standard iterator only yields online. It might be possible to merge them and still be useful. Given https://github.com/NucleoidMC/plasmid/commit/4b4815e77929a82d0fba2be03847ee9502bdd3c2, it should be possible to access everything needed through the `PlayerSet`.

Gegy: A further note- given there's quite a few `Map<GameTeam, T>` it might be useful to combine them into a single state object.

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
Patbox: Yep, it's player name and formatting is samish to one from vanilla teams

Gegy: Maybe worth renaming to like `formatPlayerName` or similar?

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
Patbox: I think it will work fine, as long as it's reapplied

src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
Patbox: I will merge these maps. But with merging refSet and PlayerSet, I'm not sure. I prefer to keep PlayerSet as it's own thing, so it's method won't apply to players that left, while still keeping reference to them

src/main/java/xyz/nucleoid/plasmid/mixin/chat/PlayerEntityMixin.java
```diff
@@ -0,0 +1,24 @@
...
+import xyz.nucleoid.stimuli.Stimuli;
+
+@Mixin(PlayerEntity.class)
+public class PlayerEntityMixin {
+
+    @Inject(method = "getDisplayName", at = @At("TAIL"), cancellable = true)
+    private void callDisplayNameEvent(CallbackInfoReturnable<Text> cir) {
+        if (((Object) this) instanceof ServerPlayerEntity player) {
+            try (var invokers = Stimuli.select().forEntityAt(player, player.getBlockPos())) {
+                cir.setReturnValue(invokers.get(GamePlayerEvents.DISPLAY_NAME).onDisplayNameCreation(player, cir.getReturnValue(), cir.getReturnValue()));
```
Gegy, line 20: Can probably avoid calling this if it's not changed- should help for mod compatibility

src/main/java/xyz/nucleoid/plasmid/util/PlayerRef.java
```diff
@@ -13,9 +13,7 @@
 public final class PlayerRef {
     private final UUID id;
 
-    private PlayerRef(UUID id) {
```
Gegy, line 16: Some odd formatting changes here 😄

PR #158, opened 2021-07-25
More SidebarWidget settings and ScrollableSidebarWidget

Patbox:


src/main/java/xyz/nucleoid/plasmid/game/common/GlobalWidgets.java
```diff
@@ -56,6 +58,74 @@ public SidebarWidget addSidebar(Text title) {
...
+        return this.addWidget(new SidebarWidget());
+    }
+
+    /**
+     * Adds a sidebar for selected players associated with this {@link GlobalWidgets} instance.
+     *
+     * @param playerChecker function returning true for players that can see this sidebar
+     * @return the created {@link SidebarWidget}
+     */
+    public SidebarWidget addSidebar(Text title, Function<ServerPlayerEntity, Boolean> playerChecker) {
```
Gegy, line 76: I think these would make sense as `Predicate<ServerPlayerEntity>`?

Patbox: Oh right, fixing it rn

Gegy (approved):

PR #162, opened 2021-08-08
Plasmid 0.5 statistics changes

Tom-The-Geek:
Changes to the statistics API found while implementing the backend.
Draft because I haven't finished the backend stuff yet.

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
+++ b/src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
@@ -43,6 +43,14 @@ public class StatisticKey<T extends Number> {
         return obj;
     }
 
+    public String getTranslationKey() {
+        return "statistic." + this.id.getNamespace() + "." + this.id.getPath();
+    }
+
+    public String getTranslationKey(String bundle) {
+        return "statistic." + bundle + "." + this.id.getNamespace() + "." + this.id.getPath();
+    }
+
     private String encodeType() {
         return this.valueType.asString() + "_" + this.storageType.asString();
     }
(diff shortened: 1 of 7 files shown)


Tom-The-Geek:
I don't think there's anything else to do here, so this is ready. I'm going to push backend stuff in a minute/maybe tomorrow

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
```diff
@@ -43,6 +43,14 @@ public JsonObject encodeValue(T value) {
         return obj;
     }
 
+    public String getTranslationKey() {
```
Gegy, line 46: Might it be a bit confusing for users of this API to know which translation key method to use?

Tom-The-Geek: Hmm, I could add a javadoc explaining the purpose of the two methods. Ideally it would be one method, but I'm not sure if there's a clean way to test if a translation key exists or not, so at the minute its left up to the caller

Tom-The-Geek: Pushed some javadoc that should explain the difference

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
```diff
@@ -43,6 +42,10 @@ public JsonObject encodeValue(T value) {
...
+        return "statistic." + this.id.getNamespace() + "." + this.id.getPath();
```
haykam821, line 46: Use `Util.makeIdentifier` here.

haykam821 (changes requested):

src/main/java/xyz/nucleoid/plasmid/game/stats/StatisticKey.java
Tom-The-Geek: I assume this is the method you meant

Gegy (approved):

PR #180, opened 2021-11-16
Update fr_fr.json

Hugman76:
# TODO
- [ ] Everything under `text.plasmid.map.`


arthurbambou:
hm, we should close this now that we have weblate

PR #196, opened 2022-01-06
Add an empty statistic bundle implementation

haykam821:
Empty statistic bundles can be used to safely handle a game configuration that does not specify a bundle namespace without `null`-checking each time the statistics bundle is accessed. This can be accomplished through the added `bundleOrEmpty` method.


ashhhleyyy (approved):

Gegy:
It seems odd to handle a lookup on null like this inside Plasmid code- does it not make more sense for the user to null-check? A null bundle namespace doesn't really "mean" anything to Plasmid.

ashhhleyyy:
The idea behind this (discussed on Discord [here](https://discord.com/channels/733462796582387753/740596279310155777/927671419884683314)) is to make it easier on the user-side code by avoiding having null-checks everywhere that the statistics are used, and move it to a single common place in plasmid code.

PR #199, opened 2022-01-07
Update WoodType.java

Shnupbups:
Now a record
Added Potted Sapling, Wood, Trapdoor, and Wall Sign
Added Crimson and Warped types

Note: breaking change as various enum methods as well as the ability to use switch are lost

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/util/WoodType.java
+++ b/src/main/java/xyz/nucleoid/plasmid/util/WoodType.java
@@ -5,36 +5,34 @@ import net.minecraft.block.Blocks;
 import net.minecraft.item.BlockItem;
 import net.minecraft.item.Item;
 import net.minecraft.item.Items;
+import org.jetbrains.annotations.Nullable;
 
-public enum WoodType {
-    OAK(Blocks.OAK_SAPLING, Blocks.OAK_LEAVES, Blocks.OAK_LOG, Blocks.OAK_PLANKS, Blocks.OAK_SLAB, Blocks.OAK_STAIRS, Blocks.OAK_FENCE, Blocks.OAK_FENCE_GATE, Blocks.OAK_DOOR, Blocks.OAK_SIGN, Blocks.OAK_BUTTON, Blocks.OAK_PRESSURE_PLATE, Items.OAK_BOAT),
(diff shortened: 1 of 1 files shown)


Gegy:
This is definitely a welcome change. If we want this to be introduced now, we should definitely keep deprecated get methods. But I'm not sure if switching from an enum is already enough of a breaking change that we will need to postpone this to 0.6 anyway?

src/main/java/xyz/nucleoid/plasmid/util/WoodType.java
```diff
@@ -5,6 +5,7 @@
 import net.minecraft.item.BlockItem;
 import net.minecraft.item.Item;
 import net.minecraft.item.Items;
+import org.jetbrains.annotations.ApiStatus;
```
haykam821, line 8: `@Deprecated(forRemoval = true)` would be better to use.

src/main/java/xyz/nucleoid/plasmid/util/WoodType.java
```diff
@@ -5,36 +5,35 @@
...
+    public static final WoodType WARPED = register(new WoodType(Blocks.WARPED_FUNGUS, Blocks.POTTED_WARPED_FUNGUS, Blocks.WARPED_WART_BLOCK, Blocks.WARPED_STEM, Blocks.WARPED_HYPHAE, Blocks.WARPED_PLANKS, Blocks.WARPED_SLAB, Blocks.WARPED_STAIRS, Blocks.WARPED_FENCE, Blocks.WARPED_FENCE_GATE, Blocks.WARPED_DOOR, Blocks.WARPED_TRAPDOOR, Blocks.WARPED_SIGN, Blocks.WARPED_WALL_SIGN, Blocks.WARPED_BUTTON, Blocks.WARPED_PRESSURE_PLATE, null));
+
+    private static final Set<WoodType> TYPES = new HashSet<>();
+    
+    private static WoodType register(WoodType woodType) {
+        TYPES.add(woodType);
+        return woodType;
+    }
+    
+    public static Set<WoodType> values() {
```
haykam821, line 31: To match enums, this should return an array rather than a set. In addition, other enum methods are missing. Of course, this is fine if you still want to introduce breaking changes.

haykam821 (changes requested):

src/main/java/xyz/nucleoid/plasmid/util/WoodType.java
Shnupbups: Added an array version, deprecated for removal. As for other methods, not gonna implement those, so this is still a breaking change, but nonetheless this method is attempting to minimize breakage.

PR #202, opened 2022-01-08
NBT Renames

### LINKED ISSUE #198: Rename usages with legacy NBT naming
FabricMC/yarn#2090 changed most NBT names, but Plasmid did not update some names accordingly:

- `ScoreStorage` methods (`toTag`/`fromTag`)
- Map packet documentation
- Various locals and parameters

Shnupbups:
Resolves #198
Note: some things need to be renamed in Map Templates

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/storage/ServerStorage.java
+++ b/src/main/java/xyz/nucleoid/plasmid/storage/ServerStorage.java
@@ -36,11 +36,11 @@ public interface ServerStorage {
      * Called when saving this {@link ServerStorage} to NBT.
      * @return - The serialized NBT data of this {@link ServerStorage}.
      */
-    NbtCompound toTag();
+    NbtCompound writeNbt();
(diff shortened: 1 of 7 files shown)


LambdAurora (approved):

src/main/java/xyz/nucleoid/plasmid/storage/ServerStorage.java
```diff
@@ -36,11 +36,27 @@
      * Called when saving this {@link ServerStorage} to NBT.
      * @return - The serialized NBT data of this {@link ServerStorage}.
      */
-    NbtCompound toTag();
+    NbtCompound writeNbt();
```
Gegy, line 39: This breaking might be unavoidable, given classes previously implementing this type would now no longer compile due to missing implementations for the new methods. We might only be able to get around it by making all the serialization functions default, but that's not ideal.

PR #218, opened 2022-03-15
Rename de_DE.json to de_de.json

arthurbambou:
datapacks don't support filename containing uppercase characters


Patbox:
I think this needs to be changed in weblate config, not here. Through @ashhhleyyy should know more

arthurbambou:
We already have looked into it, and it's not available from weblate itself yet.

ashhhleyyy:
Unfortunately, [weblate doesn't currently support this](https://github.com/WeblateOrg/weblate/discussions/7301), however renaming the file in the repo should cause it to be corrected in weblate too

PR #242, opened 2022-08-15
Add Spanish and Catalan translations

Londiuh:


Gegy:
Thank you so much for these translations! As @arthurbambou has added here, this should ideally go through [Weblate](https://hosted.weblate.org/projects/nucleoid/)

PR #245, opened 2022-09-01
Add the kick subcommand to the game command

haykam821:


src/main/java/xyz/nucleoid/plasmid/command/GameCommand.java
```diff
@@ -400,4 +406,29 @@ private static int listGames(CommandContext<ServerCommandSource> context) {
...
+
+        int successes = 0;
+
+        for (var target : targets) {
+            var gameSpace = GameSpaceManager.get().byPlayer(target);
+            if (gameSpace != null) {
+                var message = GameTexts.Kick.kick(source, target).formatted(Formatting.GRAY);
+                playerManager.broadcast(message, false);
+
+                Scheduler.INSTANCE.submit(server -> {
```
Gegy, line 424: Is the scheduler wrapping needed here? Command execution should all happen on the server thread.
Would also make it possible to use the result of `kick()` to count the number of players who were actually kicked

haykam821: I wrapped the kick in the scheduler to align with the leave subcommand.

Regarding feedback, I opted to not provide feedback to the executor because the kick messages would provide the information in most circumstances anyways. The success count provides the number of players kicked, though; is this inaccurate?

Gegy: Hmm, it's a bit weird that the leave command does that. Might be a bit of legacy behaviour from when everything was super tick-sensitive

The result isn't super important- I can't imagine the kick command being used from command blocks anyway, but it's nice to be truthful 😄

haykam821: I think the best strategy would be to leave the scheduler here and look into removing it in both places in a later pull request. That way, we can focus on testing, especially with minigames that might rely on the scheduler.

Gegy (approved):

PR #248, opened 2022-09-03
Add the saturated regeneration rule type

haykam821:
This pull request adds the saturated regeneration rule type, which can be used in conjunction with the hunger rule type to create balanced PvP games that do not use hunger.

Unfortunately, this rule type could not be implemented with the regeneration Stimuli event because that event does not currently provide enough context to differentiate saturated, peaceful, and normal natural generation.


src/main/java/xyz/nucleoid/plasmid/mixin/game/rule/HungerManagerMixin.java
```diff
@@ -0,0 +1,28 @@
...
+import net.minecraft.entity.player.HungerManager;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.server.network.ServerPlayerEntity;
+import net.minecraft.util.ActionResult;
+import xyz.nucleoid.plasmid.game.manager.GameSpaceManager;
+import xyz.nucleoid.plasmid.game.rule.GameRuleType;
+
+@Mixin(HungerManager.class)
+public class HungerManagerMixin {
+    @Inject(method = "update", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/player/PlayerEntity;heal(F)V", shift = At.Shift.BEFORE, ordinal = 0), cancellable = true)
```
haykam821, line 17: This injection point prevents the other branches of regeneration code from executing and thus will prevent the player from regenerating. The mixin added in NucleoidMC/Cake-Wars@55f903d9d1ae12d8ca420485510059b250f39dcf should be used instead.

haykam821: Fixed in 7d577c007da59c7f0c5c333e308037c881002d6e

Gegy (approved):

PR #256, opened 2022-11-29
API changes

D3lta-2-1:
Change to allow me to inject another IsolatingPlayerTeleporter in plasmid


D3lta-2-1:
this is a dumb Idea, the simplest way to don't use playerTeleporter is simply to don't call it...

PR #257, opened 2022-12-03
Abstract game space & config lists

Gegy:
A prototype at preparing for Delta's work on setting up Plasmid with a proxied server network: mostly, we need to be able to have the core `/game` commands be able to work with games that might exist remotely on another server. Currently, the `GameSpace` and `GameConfig` interface are entirely built around the idea that the games are local, but most accessors don't need that level of information. 

This PR introduces a `ListedGameSpace` and `ListedGameConfig` interface which exposes a smaller surface area for server network support. Plasmid still does not support this natively, but it at least is one step in that direction. It is however possible for external mods to in theory hook-in to add custom game space / config lists which can be used through certain commands. Ideally, in the future, we can support server network setups out-of-the-box, but that's very future work.

Not all commands will work with this setup, for example anything that operates on a player entity will naturally fail due to the player entity not existing on a remote server. Most of these I don't think are super important, but maybe some of them we would like to support in the future.


D3lta-2-1:
I think you need to know that the proxy and a server can only communicate each other be using a player connection, if there is no player on a server, it's like we don't know if it's down or not, we must use netty or faked playerConnection to communicate with other servers.

I think choosing an host server is the responsibility of the proxy, because it's him which catch any connection error. I thought plasmID as a simple slave that only process proxy requests.

But you're right, the proxy needs to know where games are instaciable, and what game is available. Concurrently they just are hardcoded in the proxy plugin.

Next, the proxy is able to handle and process directly commands, we don't need any server-side code to remake the /game command or /locate

D3lta-2-1:
but I think it will be necessary to separate the code of a gameSpace and how to join a gameSpace

PR #258, opened 2022-12-06
Features need by protein to make the direct Join

D3lta-2-1:


src/main/java/xyz/nucleoid/plasmid/game/player/PlayerOfferResult.java
```diff
@@ -60,6 +68,11 @@ public ServerWorld applyJoin(ServerPlayerEntity player) {
             return this.world;
         }
 
+        public void run()
```
haykam821, line 71: This method could possibly be called by `applyJoin`.

PR #268, opened 2023-03-21
Dynamic config registry

Gegy:
Draft for now, to be integrated with a future breaking release.


Hugman76:
We can now use Fabric API 0.86.1+ for registering dynamic registries:
```java
import net.fabricmc.fabric.api.event.registry.DynamicRegistries;
import net.minecraft.registry.Registry;
import net.minecraft.registry.RegistryKey;

public class MyRegistryKeys {
    public static final RegistryKey<Registry<Bar>> BAR = RegistryKey.ofRegistry(new Identifier("foo", "bar"));

    public static void register() {
        DynamicRegistries.register(BAR, Bar.CODEC);
    }
}
```

Hugman76:
Why are game portals still managed by their own class rather than a dynamic registry?

Gegy:
No specific reason here - definitely something that can and should be done, just needs a bit more work than configs 🙂

Hugman76:
Okay then let's leave that to another PR I think
But we might want to get it in 0.6 too

Hugman76 (approved):

Hugman76:
Also as a side note: using FAPI dynamic resistry will have to make us move all `data/games` folders to `data/plasmid/games`

PR #270, opened 2023-04-17
an attempt to avoid case where an attacker is immune to his own damage

D3lta-2-1:
add a new parameter in Team config to get rid of case where an attacker was immune to his own damage (such as TNT)


src/main/java/xyz/nucleoid/plasmid/game/common/team/TeamManager.java
```diff
@@ -328,7 +329,28 @@ private ActionResult onDamagePlayer(ServerPlayerEntity player, DamageSource sour
...
+                    || source.isOf(DamageTypes.ARROW)
+                    || source.isOf(DamageTypes.TRIDENT)
+                    || source.isOf(DamageTypes.MAGIC)
+                    || source.isOf(DamageTypes.FALLING_BLOCK)
+                    || source.isOf(DamageTypes.FALLING_STALACTITE)
+                    || source.isOf(DamageTypes.FALLING_ANVIL))
+                isProtectedFromDamageSource = !friendlyFire; //case where the attack is direct
+
+
+            if (playerTeam != null && playerTeam == attackerTeam && isProtectedFromDamageSource) {
```
Gegy, line 353: Hmm, I might not have fully understood all the context - would it make sense to instead have a check like `playerTeam == attackerTeam && attacker != player`? The `attacker` should already track indirect cause, even with things like TNT (at least in 1.19.4)

PR #299, opened 2024-07-16
Player join offer related changes

pufmat:


src/main/java/xyz/nucleoid/plasmid/game/player/GamePlayerJoiner.java
```diff
@@ -40,15 +40,9 @@ private static Set<ServerPlayerEntity> collectPlayersForJoin(ServerPlayerEntity
...
-            return results;
-        }
-
-        for (var player : players) {
-            var result = gameSpace.getPlayers().offer(player, intent);
-            if (result.isError()) {
+        var result = gameSpace.getPlayers().offer(players, intent);
+        if (result.isError()) {
+            for (var player : players) {
                 results.playerErrors.put(player, result.error());
```
Gegy, line 46: Suggestion: we can probably simplify this structure a bunch now, as we only get one error anyway :slightly_smiling_face:

src/main/java/xyz/nucleoid/plasmid/game/player/JoinOffer.java
```diff
@@ -0,0 +1,120 @@
...
+public interface JoinOffer {
+    /**
+     * @return the set of {@link GameProfile} of the players that are requesting access to this {@link GameSpace}
+     */
+    Collection<GameProfile> players();
+
+    /**
+     * @return the {@link UUID profile UUID} of the players that are requesting access to this {@link GameSpace}
+     */
+    default Collection<UUID> playerIds() {
```
Gegy, line 33: Nit: maybe a `Set`? :slightly_smiling_face:

src/main/java/xyz/nucleoid/plasmid/game/player/LocalJoinOffer.java
```diff
@@ -0,0 +1,83 @@
...
+            return this;
+        }
+
+        public void joinAll(PlayerSet players) {
+            for (var consumer : this.thenRun) {
+                consumer.accept(players);
+            }
+        }
+
+        public ServerWorld applyJoin(ServerPlayerEntity player) {
```
Gegy, line 76: I think there is some value still to running per-player logic in here, as this applies before the client is fully made aware of the teleport - so doing setup in here ensures that the state is finalised before the player sees the world again. Maybe the 'global' thenRun can still run after all players, but the per-player logic runs in here?

src/main/java/xyz/nucleoid/plasmid/game/player/JoinOffer.java
```diff
@@ -0,0 +1,120 @@
...
+     * <p>
+     * This function does not do anything on its own, but its result must be returned within a
+     * {@link GamePlayerEvents#OFFER} listener.
+     *
+     * @param world the world that the players should be teleported to when accepted
+     * @param position the position that the players should be teleported to when accepted
+     * @param yaw the 'yaw' angle that the players should be teleported to when accepted
+     * @param pitch the 'pitch' angle that the players should be teleported to when accepted
+     * @return an "accept" offer result
+     * @see JoinOfferResult.Accept#thenRun(Consumer)
```
Gegy, line 67: Nit: also link `thenRunForEach`? :slightly_smiling_face:

src/main/java/xyz/nucleoid/plasmid/game/player/JoinOffer.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    JoinOfferResult.Accept accept(ServerWorld world, Vec3d position, float yaw, float pitch);
```
Gegy, line 69: Something that I got stuck on when attempting this is that it's quite easy to unintentionally place a whole group of joining players in the same spot - even though a lot of games do place everyone in the same place - there's definitely cases where position would be randomised, or even required to be split up (e.g. in different teams). Those might even spread across separate dimensions in some cases. 
While it is technically possible to implement correctly for a game that would like to do so, it feels like an easy mistake to make. That being said, I'm not really sure what the best way to go about it is. 

It could be an extreme option to enforce that the `thenRunForEach` is responsible for all setup, but that's also error-prone in that it's not so obvious what you need to implement when creating a game.

There's also the option of having nested callbacks something like this, but I'm not hugely happy with the complexity
```java
offer.accept(acceptor -> acceptor.accept(world, randomPosition(acceptor.playerId()).thenRun(player -> ...)))
```

I think the best compromise I found is something like this:
```java
offer.accept((player, teleporter) -> {
    teleport.teleportTo(world, randomPosition(...));
    // .. do whatever else with player
})
```
It at least forces you to run logic for every player, and kind of hints that you should teleport the player by passing this 'teleporter'. We can runtime error if this function does not get called for each player.

What do you think? :slightly_smiling_face:

src/main/java/xyz/nucleoid/plasmid/game/manager/ManagedGameSpacePlayers.java
```diff
@@ -28,13 +25,23 @@ public final class ManagedGameSpacePlayers implements GameSpacePlayers {
...
-    public GameResult screenJoins(Collection<ServerPlayerEntity> players, JoinIntent intent) {
-        return this.space.screenJoins(players, intent);
+    public GameResult simulateOffer(Collection<ServerPlayerEntity> players, JoinIntent intent) {
+        if (players.stream().anyMatch(this.set::contains)) {
+            return GameResult.error(GameTexts.Join.alreadyJoined());
+        }
+
+        var offer = new LocalJoinOffer(players, intent);
+
+        return switch (this.space.offerPlayer(offer)) {
```
Gegy, line 35: Yay, this is awesome!

src/main/java/xyz/nucleoid/plasmid/game/player/GamePlayerJoiner.java
pufmat: Ok, replacing `GamePlayerJoiner.Results` with `GameResult` seems to be a good idea.

src/main/java/xyz/nucleoid/plasmid/game/player/JoinOffer.java
pufmat: Ok, I wasn't initially sure if we want `List` of `Set`.

src/main/java/xyz/nucleoid/plasmid/game/player/JoinOffer.java
pufmat: I have also found this part problematic. I added an `accept` method which only requires world, but the fact that players can be sent to different dimension makes it a bit more complicated.

In the compromise solution exposing `player` before teleporting is not a good idea because to construct player a world is required. We want this to work on multiple server instances in the future. This means that only the worlds which are available on the given server can be used, so the player can't be constructed and exposed before teleportation.

I can think of three solutions to this problem:
- The Complex Solution
Use the nested callbacks as it provides all desired functionality. It is a bit complex, but most of the time it will be only used by some kind of lobby or game abstractions. Unless we document how to use these abstractions, the offer event itself may be too hard for people who begin with creating games.
- The Lazy Solution
Just required the world and position in the `accept` method as we do now. This is not really a problem, because even if a game sends player to different dimensions, the starting lobby is probably always in the separate dimensions. However, some exotic games may not be possible or require weird implementations.
- The Clean Solution
Make offer event only decide if the players can join (remove `thenRun`), and add a new join event (actually modify as we already have it) which would be called when these players are accepted. That join event would look similar to offer event, but instead of `accept`/`reject` it would have `teleport` which works like `accept` (has to be returned from the event and provides `thenRun`).

Gegy: Ah, super good point with the player not being available at that stage! 

A potential alternative could be to have the `teleportTo` return the player instance once destination is specified. It would mean that we need to construct the player immediately after calling that function though, so how that actually is implemented with a client connecting from another server might become pretty complex. Of course fixable with another callback, but then it's become the complex solution :smile: 
It would be plausible to implement at least, so unless I'm missing something, it might be worth considering for the simplicity of usage for a game, even if the Plasmid implementation inherits a lot of the complexity.

I generally like keeping the logic of where to place the player close to the decision to allow or deny the player access - as often those would be related. Particularly, the context of all players joining might be needed to decide where to place the players (e.g. if they all get placed in a team).

pufmat: > I generally like keeping the logic of where to place the player close to the decision to allow or deny the player access - as often those would be related. Particularly, the context of all players joining might be needed to decide where to place the players (e.g. if they all get placed in a team).

I forgot to write about this, but that new join event would work on a group of players that was accepted in the offer event. Otherwise, it would end up being almost identical to the old screening and offer system. :smile: 

After sleep, I have also come up with more points for The Clean Solution.
- The split of events makes sense as the offer event must be pure, and the join event is supposed to have effects. Without this split offer event itself must be pure, but callback can have effects which may be a bit confusing.
- Fewer things can go wrong. When the join event is not implemented we can throw exception to indicate that. Teleportation in the join event can't be forgotten as the result of `teleport` must be returned from that event.

Gegy: That makes a lot of sense, let's try it! :)

src/main/java/xyz/nucleoid/plasmid/game/player/LocalJoinOffer.java
pufmat: In a different conversation we have reached a conclusion that the player is not available at this stage.

PR #318, opened 2025-01-28
Add sword blocking to old combat

haykam821:
This pull request will be part of a series of changes to expand the old combat systems to bring it further in line with Minecraft 1.8.9's combat system. As part of this pull request, old combat will now add blocking to swords, with the following behaviors:

- Blocks half damage in addition to a base of 0.5
- Does not damage the sword when blocking
- Blocks immediately, as opposed to after 0.25 seconds
- Plays damage sounds as opposed to shield sounds

The `plasmid:old_combat` data component type introduced as part of this pull request should be considered an implementation detail. Games should not expect to persist this component as its format may change in the future.


Hugman76:
I believe this is now entirely do-able with vanilla components.

PR #330, opened 2025-07-16
Feature: Allow force starting games

ellieisjelly:
Allows /game start to force start a game with a WaitingLobby even when it does not meet the minimum player requirements


Patbox:
Maybe if it required an extra argument (`force` literal) it might make sense (since it would mostly be for debug only and not production thing)

PR #333, opened 2025-08-02
Refactor registries

Hugman76:
This centralizes and exposes our registries (and values registred by Plasmid) in the API.

This is an effort to allow for further enhancements down the road, such as easier data generation, reusability and sticking to vanilla closer than before. This also set an example as to how mini-games mods should be correctly registering their content.

This PR effectively:
- Adds `PlasmidRegistries` and `PlasmidRegistryKeys` classes, containing registries and registry keys added by Plasmid.
  - Existing constants are marked as deprecated, until the next major version arrives and erases them
- Alleviates the main `Plasmid` class by registering content in dedicated API classes
  - This also exposes registered values, allowing for usages like data generation and using tags
- Deprecates existing calls to registering methods, that were located inside the registered object classes themselves
  - These methods now redirect to newer methods located in the API classes mentionned above

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/api/chat/PlasmidMessageTypes.java
+++ b/src/main/java/xyz/nucleoid/plasmid/api/chat/PlasmidMessageTypes.java
@@ -6,9 +6,9 @@ import net.minecraft.registry.RegistryKeys;
 import xyz.nucleoid.plasmid.impl.Plasmid;
 
 public final class PlasmidMessageTypes {
-    public static final RegistryKey<MessageType> TEAM_CHAT = register("team_chat");
+    public static final RegistryKey<MessageType> TEAM_CHAT = createKey("team_chat");
 
-    private static RegistryKey<MessageType> register(String key) {
+    private static RegistryKey<MessageType> createKey(String key) {
         return RegistryKey.of(RegistryKeys.MESSAGE_TYPE, Plasmid.id(key));
     }
 }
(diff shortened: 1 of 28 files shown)


src/main/java/xyz/nucleoid/plasmid/api/game/GameType.java
```diff
@@ -21,13 +22,17 @@
  * @see GameConfig
  */
 public final class GameType<C> {
-    public static final TinyRegistry<GameType<?>> REGISTRY = TinyRegistry.create();
+    /**
+     * @deprecated Use {@link PlasmidRegistries#GAME_TYPE} instead.
+     */
+    @Deprecated
+    public static final Registry<GameType<?>> REGISTRY = PlasmidRegistries.GAME_TYPE;
```
Patbox, line 29: This isn't binary compatible! It might be better to create stub TinyRegistry variant that just wraps the real one

Patbox (changes requested):

src/main/java/xyz/nucleoid/plasmid/api/game/GameType.java
Hugman76: Ah yes, that went over my head. new solution should fix this

src/main/java/xyz/nucleoid/plasmid/api/chat/PlasmidMessageTypes.java
```diff
@@ -3,9 +3,12 @@
...
 import net.minecraft.registry.RegistryKey;
 import net.minecraft.registry.RegistryKeys;
-import net.minecraft.util.Identifier;
 import xyz.nucleoid.plasmid.impl.Plasmid;
 
 public final class PlasmidMessageTypes {
-    public static final RegistryKey<MessageType> TEAM_CHAT = RegistryKey.of(RegistryKeys.MESSAGE_TYPE, Identifier.of(Plasmid.ID, "team_chat"));
+    public static final RegistryKey<MessageType> TEAM_CHAT = register("team_chat");
+
+    private static RegistryKey<MessageType> register(String key) {
```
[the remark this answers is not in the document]
Hugman76: Very true, that's a better name

PR #334, opened 2025-08-02
Add team list providers

Hugman76:
This adds all classes needed for team list providers, a way to randomly generate a `GameTeamList`.
A team list provider only requires a `Random` to provide a `GameTeamList`.

It also has builtin compatibility for files that are configurated using a fixed list of game teams such as `GameTeamList#CODEC`/`GameTeam#CODEC#asList()` and integer values describing team amounts such as `IntProvider#CODEC`/`Codec#INT`.

## Types
The registry for types is `plasmid:team_list_provider_type` (found in `PlasmidRegistries#TEAM_LIST_PROVIDER_TYPE`). Obviously, any mod can add to it.

Here are the different team provider types:
- `plasmid:constant` - a fixed list of `GameTeam`.
- `plasmid:trim` - trims another provider to a certain size
  - Can be also be shuffled before the trim for more variety
- `plasmid:random` - picks from a list of other providers.
- `plasmid:sized_alternatives` - picks for a certain size, and then picks from a map of sizes to providers.
  - The map field is optional, and if left empty will pick up from a standard map set up to support from 1 up to 16 teams. (see `DefaultTeamLists`) This default map provides handpicked basic teams with constrated and vivid colors. The lesser teams they are, the more effective it is at constrasting everything.

## Codec
The base Codec (`TeamListProvider#CODEC`) supports many formats:
- A typed provider, which is the base format
- A list of game teams. Instead of using `plasmid:constant`, you can directly pass the fixed list. This means existing configurations that use `GameTeamList#CODEC` can be replaced with the base codec and generated configurations will have the same functionality.
- An integer provider (which can be inlined as a constant int).  This replaces the usage of `plasmid:sized_alternatives` with the default map.

If you are using data generation, please check around the `TeamListProvider` classes for helper functions, noticeably some that use the default team map.

## Other details
You can try out team providers in the dev environment by launching the test mod.

If we are discontent with the teams provided by default, we will always be able to change them in the future and still maintain perfect compatiblity with previous datapack resources.


src/main/java/xyz/nucleoid/plasmid/api/game/common/team/provider/TeamListProvider.java
```diff
@@ -0,0 +1,53 @@
...
+import xyz.nucleoid.plasmid.api.registry.PlasmidRegistries;
+
+/**
+ * Provides a {@link GameTeamList}.
+ *
+ * @see TeamListProviderType
+ *
+ * @author Hugman
+ */
+public abstract class TeamListProvider {
```
Patbox: Would it make more sense for this to be an interface? Since while it's abstract, it has no methods, fields or constructors on it's own. Other TeamListProvider implementations are record like, not being record purely due to this being a class

src/main/java/xyz/nucleoid/plasmid/api/game/common/team/provider/TeamListProviderType.java
```diff
@@ -0,0 +1,27 @@
...
+import net.minecraft.util.Identifier;
+import xyz.nucleoid.plasmid.api.registry.PlasmidRegistries;
+import xyz.nucleoid.plasmid.impl.Plasmid;
+
+/**
+ * Types of {@link TeamListProvider} that can be registered in the registry ({@link PlasmidRegistries#TEAM_LIST_PROVIDER_TYPE}).
+ *
+ * @author Hugman
+ */
+public record TeamListProviderType<T extends TeamListProvider>(MapCodec<T> codec) {
```
Patbox: While not sure if it would be better or not (or if you have future plans), does it make sense to wrap the codec? Vanilla in many cases just uses codec directly as the type (through not all cases).

Patbox:
Aside it not compiling now, LGTM

Hugman76:
Oops, here we go

PR #335, opened 2025-08-05
Add map template processors

Hugman76:
This adds all classes needed for map template processors, a way to modify dynamically a `Template`.
A map template processor requires a given `Template` and `GameActivity` it is fired from.

Since it needs a GameActivity, that means in **most cases** the game opening sequence requires to set up the activity before the template, and therefore before the world is generated (because the world requires the template for its chunk generator most of the time)

It might look barebones right now, since it only adds 3 processor types that can't be configured a lot, but this is meant to evolve greatly in the future (with set bounds configuration for most processors, for example). This is aiming to be tested in various games.

## Types
The registry for types is `plasmid:map_template_processor_type` (found in `PlasmidRegistries#MAP_TEMPLATE_PROCESSOR_TYPE`). Obviously, any mod can add to it.

Here are the different map template processors:
- `plasmid:replace_blocks` - maps blocks, and tries to keep states in between.
- `plasmid:replace_block_entities` - maps strings to search and replace in block entity NBT data.
- `plasmid:team_colors` - dynamically generates processors of both previous types for a given set of colors to be remapped to team colors. This processor needs to know about the game's team list, hence the addition of `PlasmidGameAttachments#TEAM_LIST` that must be attached to the game activity to be used.


Patbox:
Looks good, through the general types / structure would maybe make more sense in Map Template library itself? Through it would need bit differently setup contexts hm. Otherwise LGTM

Hugman76:
I'm all for moving this work to the library itself, but like you said it needs a different context set up, and that's because of the GameActivity required for the team_colors processor.
I could try out having a parameter map like Minecraft has for loot? We'd have more flexibility in which contexts we use map template processors too.

Hugman76:
Aside, I'd be grateful if we took the time to port it to the library later than this PR, for testing purposes

Hugman76:
Updated the PR with the parameters map I was talking about earlier.

Patbox:
Ok decided it makes sense to keep it in plasmid for now, and migrate once some more work is done in templates library.

PR #339, opened 2025-08-15
Allow constructing template chunk generators without a server instance

haykam821:
Data generators are [serverless](https://miro.medium.com/v2/format:webp/1*DgdXbb-mdbrKE_J9zsSQXw.jpeg) infrastructure, so having a `TemplateChunkGenerator` constructor that uses a `RegistryWrapper.WrapperLookup` parameter would be useful.

### DIFF
--- a/src/main/java/xyz/nucleoid/plasmid/api/game/world/generator/GameChunkGenerator.java
+++ b/src/main/java/xyz/nucleoid/plasmid/api/game/world/generator/GameChunkGenerator.java
@@ -2,6 +2,7 @@ package xyz.nucleoid.plasmid.api.game.world.generator;
 
 import net.minecraft.registry.RegistryKey;
 import net.minecraft.registry.RegistryKeys;
+import net.minecraft.registry.RegistryWrapper;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.ChunkRegion;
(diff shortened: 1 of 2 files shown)


Gegy (approved):

Gegy:
Don’t see why not, and code looks good - but a bit confused about data generation since we don’t have Codecs for these at all iirc 😅 (should we?)

haykam821:
You're right that there isn't an existing codec for `TemplateChunkGenerator`. Since map templates are usually loaded from the `MapTemplateSerializer.loadFromResource` method which itself requires a server instance, I'm not sure if there is a good way to serialize `TemplateChunkGenerator` as is. We don't get the map template ID as context in this chunk generator either, and map templates might be generated dynamically and not have IDs anyways.

In the latter case, then the chunk generator and generated map template can be represented by a subclass that provides its own codec. A `CubeChunkGenerator` might provide a codec that takes size and block state provider parameters, for example. Thinking about it, though, maybe a constructor that accepts a `RegistryEntryLookup<Biome>` would make more sense to allow the `RegistryOps.getEntryLookup` method to be used in said codecs.

As an aside, this chunk generator should probably be moved to the Map Templates library at some point. I know there's also some work being done with [map template processors](https://github.com/NucleoidMC/plasmid/pull/335) which might make sense as a general set of improvements to how map templates are loaded and configured. It would be nice to get rid of the code pattern and corresponding translation that is used in nearly every game that loads maps:

https://github.com/NucleoidMC/Paintball/blob/a24647c9c6d948529a16621e5f251aea511b5ecf/src/main/java/io/github/haykam821/paintball/game/map/PaintballMapBuilder.java#L19-L26

PR #341, opened 2025-09-09
Add vote starting to the Waiting Lobby

ellieisjelly:
This is to eliminate lengthy waits for e.g when there is all but one player in the game lobby


ellieisjelly:
I do want some feedback though: should we allow players to start the game even if there aren't enough people to meet the minimum set in PlayerLimiter? The current behaviour allows this but specifies that you'd be playing under the recommended amount.

Hugman76:
I believe the minimum field marks the very bare minimum amount of players in order to have a game start up and be finishable with that amount of players. The threshold value is present for the intended, more realistic minimum

ellieisjelly:
> I believe the minimum field marks the very bare minimum amount of players in order to have a game start up and be finishable with that amount of players. The threshold value is present for the intended, more realistic minimum

I often see the threshold used as a recommended amount though, maybe this feature could be a config option you can set in the waiting lobby config?

Hugman76:
I think we are describing the same usage for the threshold value. I think having a minimum, a threshold and a maximum value is already enough.
To me, the voting should be available as long as `/game start` successfully starts the game, which to me is synonym of having crossed the minimum amount.

ellieisjelly:
> I think we are describing the same usage for the threshold value. I think having a minimum, a threshold and a maximum value is already enough. To me, the voting should be available as long as `/game start` successfully starts the game, which to me is synonym of having crossed the minimum amount.

New commit addresses this

ellieisjelly:
This PR is ready for review

ZekeZDev:
image
Seems that the red vote start wool is kept to player inventory

this was after starting game btw

after looting it was removed so I think its some kinda desync

ZekeZDev:
Also what is the required tags for allowing a game to have this, uhc, deathrun and a couple others dont have it.

They dont have the item but they do still have the text:
image
*picture from UHC game, no vote item in hand*

ZekeZDev:
I do also think that https://github.com/FoundationGames/Builderdash 's implementation with the bossbar and such has better user feedback and also personally prefer the playerhead. I do think this is a good start to universal but I wonder if some of the qol from that could be used. (AFAIK, they are license compatible)

ellieisjelly:
> image Seems that the red vote start wool is kept to player inventory
> 
> this was after starting game btw
> 
> after looting it was removed so I think its some kinda desync

is this happening on skywars? if so, this is a regression caused by a previous plasmid commit and is already referenced under NucleoidMC/skywars#57

ellieisjelly:
> Also what is the required tags for allowing a game to have this, uhc, deathrun and a couple others dont have it.
> 
> They dont have the item but they do still have the text: picture from UHC game, no vote item in hand

The voting wool is only available when the minimum number of players are met, though the voting text shouldnt be showing up so I'll fix that

ZekeZDev:
> > image Seems that the red vote start wool is kept to player inventory
> > this was after starting game btw
> > after looting it was removed so I think its some kinda desync
> 
> is this happening on skywars? if so, this is a regression caused by a previous plasmid commit and is already referenced under [NucleoidMC/skywars#57](https://github.com/NucleoidMC/skywars/issues/57)

Ah yep it would be that regression.

ZekeZDev:
> > Also what is the required tags for allowing a game to have this, uhc, deathrun and a couple others dont have it.
> > They dont have the item but they do still have the text: picture from UHC game, no vote item in hand
> 
> The voting wool is only available when the minimum number of players are met, though the voting text shouldnt be showing up so I'll fix that

I do still think the message should be action bar as chat often gets overlooked.

ellieisjelly:
> I do still think the message should be action bar as chat often gets overlooked.

This is now the case

PR #343, opened 2025-09-29
Update mod to Minecraft 1.21.9

Hugman76:
Compiles against 1.21.9, with fixed mixins

Stimuli has at least one blocking mixin


src/main/java/xyz/nucleoid/plasmid/mixin/game/space/PlayerManagerMixin.java
```diff
@@ -137,8 +143,6 @@ private void respawnPlayer(
         }
 
         player.setServerWorld(world);
-
-        player.readGameModeData(playerData);
```
Hugman76, line 141: Please note this line is no longer needed as it is part of `ServerPlayerEntity#readData` now. The previous method would do nothing if the data was null.

